PR #1, opened 2019-09-26
Add Rails

denny:


config/master.key
```diff
@@ -0,0 +1 @@
+d1399141ddd73ff3e240dce1c035ebc3
```
eliotsykes: This file looks sensitive and that it shouldn't be tracked by git, especially not in a public repo. 

Double-check this is correct, and if so, treat any credentials encrypted in `config/credentials.yml.enc` as exposed and invalidate/reset them. Also generate a fresh `master.key`, remove it from git permanently, and re-encrypt `config/credentials.yml.enc`.

denny: D'oh! Thanks.

denny: I didn't know either of these files existed, so I don't think there's anything of consequence in them?

eliotsykes: Try decrypting it to check:
https://github.com/rails/rails/blob/master/railties/lib/rails/commands/credentials/USAGE

If there's any secrets in there, such as the one used to encrypt cookies, then that will need regenerating. Also regenerate `master.key` as its exposed, even if nothing sensitive is in `credentials.yml.enc` yet.

eliotsykes: Try running: `EDITOR="atom --wait" bin/rails credentials:edit`

denny: Ah yes, the cookie key is in it (nothing else though). Thanks again!

denny: I've gone for the brute force approach of removing the files, and adding SECRET_BASE_KEY to .env instead

PR #6, opened 2019-10-02
Bugfixes and improvements for pages feature

denny:


app/views/admin/pages/sections/edit.html.erb
```diff
@@ -1,6 +1,6 @@
 <% @page_title = 'Edit section' %>
 
-<%= form_for( @section, url: "/admin/page/section/#{@section.id}", method: 'post' ) do |f| %>
+<%= form_for @section, url: "/admin/pages/section/#{@section.id}", method: 'post' do |f| %>
```
eliotsykes: Prefer to use route helper methods instead of hard-coding the paths, e.g.

```diff
-<%= form_for @section, url: "/admin/pages/section/#{@section.id}", method: 'post' do |f| %>
+<%= form_for @section, url: admin_pages_section_path(@section), method: 'post' do |f| %>
```

`config/routes.rb` might need tweaking to support the above method. Run `rails routes` to determine the correct `*_path` method name.

config/routes.rb
```diff
@@ -22,18 +22,18 @@
...
-      get  'sections',     to: 'sections#index'
-      get  'sections/add', to: 'sections#new'
-      post 'sections/add', to: 'sections#create'
-      get  'sections/:id', to: 'sections#edit'
-      post 'sections/:id', to: 'sections#update'
+      get  'sections',    to: 'sections#index'
+      get  'section/add', to: 'sections#new'
+      post 'section/add', to: 'sections#create'
+      get  'section/:id', to: 'sections#edit'
+      post 'section/:id', to: 'sections#update'
```
eliotsykes: Consider using Rails' "resourceful" routes, they generate reasonable default paths+HTTP Methods, along with good names for helper `*_url/*_path` methods:

```rb
namespace :pages do
  resources :sections, only: [:index, :new, :create, :edit, :update]
  resources :templates, only: [:index, :new, :create, :edit, :update]
end
```

denny: I did initially go down that route (heh) but ended up backing it out in favour of hand-rolling the URLs. I can't remember for sure, but I _think_ the problem was that they pluralised 'sections' on the pages that handle a single section (case in point, the edit page).

I'll have another look, but if that was it then I may possibly choose to die on this particular hill rather than be annoyed every time I look at the address bar in the admin section :stuck_out_tongue:

app/views/admin/pages/sections/edit.html.erb
denny: It doesn't seem to have one. Probably because [ see below ].

denny: How about this style?

`<%= form_for( @page, url: url_for( controller: '/admin/pages', action: 'edit', id: @page.id ), method: 'post' ) do |f| %>`

config/routes.rb
eliotsykes: Got it. They can be customised using the `path` and `path_names` options, but won't quite match what you've got now. In case its useful:

```
resources :sections, path: 'section', only: [:index, :new, :create, :edit, :update]

# or

resources :sections, path: 'section', path_names: { edit: '', new: 'add' }, only: [:index, :new, :create, :edit, :update]
```

app/views/admin/pages/sections/edit.html.erb
eliotsykes: > It doesn't seem to have a suggested path for any of the edit methods

Ah right, add an `as: :some_name` option to the route to generate a named helper method, e.g.

```rb
get  'section/:id', to: 'sections#edit', as: :section
```

This will generate an `admin_pages_section_path` helper.

denny: Oooh. Shiny. :)

config/routes.rb
denny: I took a swing at this today, but in hindsight the outcome was entirely predictable: Rails doesn't like it when you don't follow the One True Way. And it will punish you.

I ended up with so much gaffa tape over special cases that it's far messier than sticking with the individual routes; in particular, it'll be way harder to follow for any hypothetical new people coming into the project.

I think I'll change my hand-rolled URLs from 'add' to 'new' to follow convention a bit more, and other than that treat this as a painful learning experience :grimacing: :grinning:

denny: https://github.com/denny/ShinyCMS-ruby/commit/91704d0618ee1142b8cf6e762ade78ee012b4cff is the main commit on the experimental branch, although there's a bunch of unrelated stuff in there too, it hasn't been a very focused weekend :)

PR #7, opened 2019-10-02
Add tests for page display on main site, view hidden pages in admin area

denny:
The continuing adventures of the pages feature-set.


app/models/page.rb
```diff
@@ -16,6 +16,10 @@ def self.top_level_hidden_pages
     Page.where( section: nil, hidden: true )
   end
 
+  def self.are_there_any_hidden_pages?
+    Page.where( hidden: true ).count.positive?
```
eliotsykes: `exists?` also works for these kind of queries:

```diff
-Page.where( hidden: true ).count.positive?
+Page.exists?(hidden: true)
```

https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-exists-3F

app/views/admin/pages/index.html.erb
```diff
@@ -44,3 +44,57 @@
...
+<table class="list">
+	<tr>
+		<th class="main">
+			Page
+		</th>
+		<th>
+			Actions
+		</th>
+	</tr>
+	<% if @tl_hidden_pages.count.positive? %>
```
eliotsykes: Admittedly this is a minor premature optimization below, but you might also prefer the last option using `#any?` for readability, and treat the optimization as a bonus:

```rb
# #count will *always* perform a SQL COUNT query:
<% if @tl_hidden_pages.count.positive? %>
```
```rb
# Using #size instead of #count does one of:
#
# 1) If @tl_hidden_pages is NOT loaded (i.e. no SELECT * query done yet),
# then a SQL COUNT query is performed.
#
# 2) If @tl_hidden_page IS loaded (i.e. SELECT * query done),
# then no SQL COUNT query is performed, instead @tl_hidden_pages.length
# (Array#length) is called, saving a trip to the DB.
<% if @tl_hidden_pages.size.positive? %>
```
```rb
# Using #any? instead performs a similar optimization to
# #size doing one of:
#
# 1) If @tl_hidden_pages is NOT loaded (i.e. no SELECT * query done yet),
# then a SQL SELECT query is performed.
#
# 2) If @tl_hidden_page IS loaded (i.e. SELECT * query done),
# then no SQL SELECT query is performed, instead !@tl_hidden_pages.empty?
# (!Array#empty?) is called, saving a trip to the DB.
# 
# (also shortens the method chain by one which is nice for readability)
<% if @tl_hidden_pages.any? %>
```

If you ever want to experiment to see the differences in what SQL is run:

```rb
bin/rails console

# Enable SQL logging in console:
ActiveRecord::Base.logger = Logger.new(STDOUT); nil

# Need `; nil` on end of line to prevent console from calling tl_hidden_pages.inspect
# and prematurely querying the DB:
tl_hidden_pages = Page.top_level_hidden_pages; nil

tl_hidden_pages.count # should perform SQL COUNT

tl_hidden_pages.size # should perform SQL COUNT

tl_hidden_pages.any? # should perform SQL SELECT

# Call a method that would load tl_hidden_pages and perform SELECT * query,
# e.g.:
tl_hidden_pages.load
# tl_hidden_pages.each(&:id) # looping would also perform SELECT * query.

tl_hidden_pages.size # NO SQL COUNT query
tl_hidden_pages.any? # NO SQL SELECT query
tl_hidden_pages.count # performs (unnecessary) SQL COUNT query
```

spec/rails_helper.rb
```diff
@@ -1,12 +1,17 @@
...
 ENV['RAILS_ENV'] ||= 'test'
 require File.expand_path('../config/environment', __dir__)
+
 # Prevent database truncation if the environment is production
 abort 'Rails is running in production mode!' if Rails.env.production?
+
 require 'rspec/rails'
 # Add additional requires below this line. Rails is not loaded until this point!
 
+require 'support/factory_bot'
```
eliotsykes: Consider removing this line 13 and uncommenting  line 28 so all files under `spec/support` are required.

spec/support/factory_bot.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.configure do |config|
+  config.include FactoryBot::Syntax::Methods
```
eliotsykes: This config means FactoryBot's methods like `create` are available in all specs without prefixing with `FactoryBot`, for example:

```diff
-page = FactoryBot.create :top_level_page
+page = create :top_level_page
```

denny: Apparently I realised that halfway through the spec file and didn't go back to fix the earlier ones. How odd. Cheers! :)

app/views/admin/pages/index.html.erb
denny: Neat!

PR #12, opened 2019-10-04
CSS => SCSS, Page admin feature fixes

denny:
Move admin area to SCSS for great justice / colour name variables, and fix the drop-downs in the admin edit pages to reselect the current setting. And stuff.


app/views/admin/pages/edit.html.erb
```diff
@@ -19,13 +19,13 @@
   </p>
   <p>
     Template:
-    <br><%= select( 'page', 'template_id',
-      PageTemplate.all.collect { |t| [ t.name, t.id ] } ) %>
+    <br><%= select( 'page', 'template_id', PageTemplate.pluck( :name, :id ),
```
eliotsykes: Consider using the form builder (`f`) `select` method, and it ought to set the `selected` option correctly:

```erb
<%= f.select(:template_id,  PageTemplate.pluck( :name, :id ) %>
```

app/views/admin/pages/edit.html.erb
```diff
@@ -19,13 +19,13 @@
...
+            { selected: @page.template_id} ) %>
   </p>
   <p>
     Section:
-    <br><%= select( 'page', 'section_id',
-      PageSection.all.collect { |s| [ s.name, s.id ] }, { include_blank: 'None' } ) %>
+    <br><%= select( 'page', 'section_id', PageSection.pluck( :name, :id ),
```
eliotsykes: As above, try form builder `select` method: `f.select(...)`

eliotsykes: Also consider using `f.label(:section_id, "Section")` for a11y (line 26).

app/views/admin/pages/index.html.erb
```diff
@@ -36,13 +33,63 @@
...
       <%= link_to 'View', url_for( "/pages/#{section.slug}/#{page.slug}" ) %>
-      <%= link_to 'Edit', url_for( controller: '/admin/pages', action: 'edit', id: page.id ) %>
+      <%= link_to 'Edit', url_for( controller: '/admin/pages', action: 'edit', id: page ) %>
 		</td>
 	</tr>
-	<% end %>
-	<% end %>
+
+  <% indent = indent + 2 %>
+  <% section.sections.each do |subsection| %>
```
eliotsykes: The table row/cell generation code looks like a good candidate for DRYing up with partial template(s).

app/views/admin/pages/edit.html.erb
denny: Aha, cool, _much_ neater.

What you see here is what I found when googling for syntax, I did wonder why it wasn't an f.* like the rest.

app/views/admin/pages/index.html.erb
denny: Yeah, particularly as it's a recursive data structure ("It's turtles all the way down!")

denny: Boom! https://github.com/denny/ShinyCMS-ruby/pull/35/commits/5aaf4715161b8aced8c1d2f93956543c2b8560c8

PR #14, opened 2019-10-04
Add test for site settings get() method

denny:


app/models/site_setting.rb
```diff
@@ -1,6 +1,6 @@
 # Model class for site settings
 class SiteSetting < ApplicationRecord
   def self.get( name )
-    find_by( name: name )
+    find_by( name: name ).value
```
eliotsykes: `Setting.get(...)` seems like the kind of method that will be called many times per request when generating a response. 

Consider reducing the resources it uses by making use of `.pluck(:value)`.

```diff
-find_by( name: name ).value
# Avoids loading a Setting object, and only pulls
# back a single column from DB.
+where(name: name).pluck(:value)
```

denny: This brought back an array (containing one element), so I added .first to it.

This doesn't throw a nil error for .pluck or .first even if the setting doesn't exist; I assume .pluck must be returning an empty array rather than nil for .first to work, is it doing that because .where returns an empty collection if there are no matching rows, whereas find_by was returning a nil?

eliotsykes: > This doesn't throw a nil error for .pluck or .first even if the setting doesn't exist; I assume .pluck must be returning an empty array rather than nil for .first to work, is it doing that because .where returns an empty collection if there are no matching rows, whereas find_by was returning a nil?

Yep!

> This brought back an array (containing one element), so I added .first to it.

Nice, I missed that. Rails 6 added a `pick` method which covers this:

```diff
-where(name: name).pluck(:value).first
+where(name: name).pick(:value)
```

https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-pick

PR #17, opened 2019-10-05
Site settings page in admin area working for add/delete

denny:


app/controllers/admin/settings_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  def index
+    @settings = Setting.order( :name )
+  end
+
+  # New setting form submitted
+  def create
+    @setting = Setting.new( setting_params )
+
+    if @setting.save
+      flash[ :notice ] = 'New setting added'
```
eliotsykes: If the CMS ought to support i18n in the near future, consider adding any new flash messages to `config/locales/en.yml` and use `I18n.t(...)` in the controllers.

app/views/admin/settings/index.html.erb
```diff
@@ -0,0 +1,27 @@
+<% @page_title = 'Site Settings' %>
+
+<%= form_with url: admin_settings_path, method: 'post' do %>
+  <% @settings.each do |s| %>
+  <p>
+    <label class="setting_name"><%= s.name %>
+    <%= text_field :setting, s.id, value: s.value %>
+    <%= link_to 'Delete', admin_setting_path( s ) %>
```
eliotsykes: For destructive actions like this Delete link, avoid `GET` HTTP method to:

1. mitigate accidental deletes by link-preloaders and
2. enable CSRF protection

```diff
-<%= link_to 'Delete', admin_setting_path( s ) %>
+<%= link_to 'Delete', admin_setting_path(s), method: :delete %>
```
The associated route will also need to change from `get ...` to `delete ...`.

Also consider moving the delete link to outside the `<label>` element.

denny: Is there also a way to trigger an 'are you sure?' popup from the link_to tag? The code I originally had here had method: :delete as well as something to do the 'are you sure' bit, but as the latter didn't work I removed both on the grounds that I had cargo-culted them from the same place :)

eliotsykes: The `confirm: 'question?'` option covers this, see API doc: https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to

denny: After a chunk of reading around this evening, I figured out why the :delete method didn't work for me the first time I tried it; it's dependent on having jquery-ujs available, and (afaict) that's dependent on having webpack set up. I've got it working now but it does seem to have added quite a few things to the project. Hopefully they'll be useful for more than just this one thing...

eliotsykes: Including UJS in the app will become more useful as the app grows.

See https://github.com/denny/ShinyCMS-ruby/pull/26#discussion_r332875471

PR #18, opened 2019-10-05
default_page and default_section

denny:
... and more use of _path helpers in menus.


app/models/page.rb
```diff
@@ -20,15 +20,23 @@ def self.are_there_any_hidden_pages?
...
+    return page if page
 
     return Page.top_level_pages.first if Page.top_level_pages.first
 
     Page.first
   end
+
+  def self.find_default_page
+    name = Setting.get 'default_page'
+    Page.where( name: name ).or( Page.where( slug: name ) ).first
```
eliotsykes: Consider raising an error if there is more than one matching `Page`, it seems like an invalid state that would allow inconsistent behaviour.

app/models/page.rb
```diff
@@ -20,15 +20,23 @@ def self.are_there_any_hidden_pages?
...
+    return page if page
+
+    section = PageSection.default_section
+    page    = section&.pages&.default_page
+    page  ||= section&.pages&.first
+    return page if page
 
     return Page.top_level_pages.first if Page.top_level_pages.first
 
     Page.first
```
eliotsykes: Consider if this should exclude hidden pages (line 35).

app/models/page.rb
eliotsykes: Should `find_default_page` exclude hidden pages?

app/models/setting.rb
```diff
@@ -1,6 +1,6 @@
 # Model class for site settings
 class Setting < ApplicationRecord
   def self.get( name )
-    find_by( name: name ).value
+    find_by( name: name )&.value
```
eliotsykes: Using `where(...).pluck(:value)` ought to avoid needing the safe navigation operator.

PR #19, opened 2019-10-06
Tests and validations for pages/sections/templates

denny:
"Safety third!"

### DIFF
--- a/app/controllers/admin/pages/sections_controller.rb
+++ b/app/controllers/admin/pages/sections_controller.rb
@@ -31,7 +31,8 @@ class Admin::Pages::SectionsController < AdminController
     # Save edited page section details
     @section = PageSection.find( params[:id] )
 
-    if @section.update!( section_params )
+    _unused = @section.update( section_params )
+    if @section.valid?
       flash.now[ :notice ] = 'Section details updated'
     else
       flash.now[ :alert  ] = 'Failed to update section details'
--- a/app/models/page_template.rb
+++ b/app/models/page_template.rb
@@ -1,6 +1,18 @@
 # Model for page templates
 class PageTemplate < ApplicationRecord
-  # Associations
+  validates :name,     presence: true
+  validates :filename, presence: true
+
   has_many :pages, foreign_key: 'template_id', inverse_of: 'template',
                    dependent: :delete_all
+
+  # Instance methods
+
+  def file_exists?
+    File.exist?(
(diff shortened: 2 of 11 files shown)


app/controllers/admin/pages/sections_controller.rb
```diff
@@ -31,7 +31,8 @@ def update
     # Save edited page section details
     @section = PageSection.find( params[:id] )
 
-    if @section.update!( section_params )
+    _unused = @section.update( section_params )
+    if @section.valid?
```
eliotsykes: Apologies if you're already aware of this. `@section.update( section_params )` (no bang) returns the same boolean result as `@section.valid?`. 

This could be tweaked to, and is a little more conventional:

```diff
-_unused = @section.update( section_params )
-if @section.valid?
+if @section.update( section_params )
```

app/models/page_template.rb
```diff
@@ -1,6 +1,19 @@
...
   has_many :pages, foreign_key: 'template_id', inverse_of: 'template',
                    dependent: :delete_all
+
+  # Instance methods
+
+  # Check whether the template file is present on disk
+  def file_exists?
+    File.exist?(
+      Rails.root.join(
+        'app', 'views', 'pages', 'templates', "#{filename}.html.erb"
```
eliotsykes: I suspect there may be a vulnerability here that allows file traversal if the submitted filename `contains` "../../".

Consider applying a strict safelist, where only the filenames in `app/views/pages/templates/*.html.erb` are allowed, rejecting anything that doesn't match one of those filenames exactly.

`Dir.glob` is one option for implementing this: https://ruby-doc.org/core-2.6.3/Dir.html#method-c-glob

app/controllers/admin/pages/sections_controller.rb
denny: This is how my code started out, but I think some combination of my tests and rubocop wasn't happy with it. I'll have another look at some point :)

eliotsykes: It could be that the original used `update!` which will raise an error for an invalid object, meaning the `else` branch would never run, whereas `update` (no bang) will return false for an invalid object.

denny: Just reverted this code and it works fine now, so ... I guess I was tired and doing something incredibly stupid the first time :D

PR #22, opened 2019-10-06
More wrestling with helpers and URLs - mostly changing 'add' to 'new'

denny:


app/views/admin/pages/new.html.erb
```diff
@@ -1,41 +1,41 @@
...
     <br><%= f.text_field :title %>
   </p>
   <p>
-    Slug:
+    <%= f.label :slug %>
     <br><%= f.text_field :slug %>
   </p>
   <p>
-    Template:
+    <%= f.label :template %>
```
eliotsykes: I'm not sure if the `<label>` `for` attributes generated on lines 21 and 26 will be correct. Clicking on the labels might not focus the associated `<select>`. It may need to be `<%= f.label :template_id, "Template" %>`.

app/views/admin/pages/templates/new.html.erb
```diff
@@ -1,19 +1,20 @@
...
   </p>
   <p>
-    Description:
+    <%= f.label :description %>
     <br><%= f.text_field :description %>
   </p>
   <p>
-    Template filename:
-    <br><%= f.text_field :filename %> <%# TODO: replace with select drop-down %>
+    <%# TODO: replace with select drop-down %>
```
eliotsykes: Agreed 👍 . Also the submitted value will want to be strictly validated against a safelist of known filenames when saved.

This is to mitigate file traversal attack shenanigans, say if a malicious user tried to submit a filename like `../../../top/secret.txt`. And this is later paired with a separate vulnerability to reveal the contents of that file.

denny: How's this? https://github.com/denny/ShinyCMS-ruby/commit/597e8c93663af4

eliotsykes: Looks solid 👍

PR #23, opened 2019-10-06
Add spec, fiddle with routing some more

denny:


app/views/admin/settings/index.html.erb
```diff
@@ -1,20 +1,19 @@
...
+<%= form_with name: 'all_settings', url: admin_settings_path, method: :post do %>
   <% @settings.each do |s| %>
   <p>
-    <label class="setting_name"><%= s.name %>
-    <%= text_field :setting, s.id, value: s.value %>
-    <%= link_to 'Delete', admin_setting_path( s ) %>
-    </label>
+    <label for="shinycms_setting_<%= s.id %>"><%= s.name %></label>
+    <%= text_field :settings, "shinycms_setting_#{s.id}", value: s.value %>
+    <%= link_to 'Delete', admin_setting_delete_path( s ) %>
```
eliotsykes: Ah I see you've moved the link out of the label already. Feedback about using `method: :delete` still applies (CSRF  protection, mitigate accidental deletes from link preloaders).

PR #24, opened 2019-10-07
More efficient .get(), less dangerous .delete()

denny:
* Create less overhead in the .get method, by using .pluck() instead of .find_by().value
* Prevent link preloaders from causing setting-removal chaos, by using DELETE instead of GET for the delete link.

### DIFF
--- a/app/views/admin/settings/index.html.erb
+++ b/app/views/admin/settings/index.html.erb
@@ -4,7 +4,7 @@
   <% @settings.each do |s| %>
   <p>
     <label for="shinycms_setting_<%= s.id %>"><%= s.name %></label>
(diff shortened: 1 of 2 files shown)


app/views/admin/settings/index.html.erb
```diff
@@ -4,7 +4,7 @@
   <% @settings.each do |s| %>
   <p>
     <label for="shinycms_setting_<%= s.id %>"><%= s.name %></label>
-    <%= text_field :settings, "shinycms_setting_#{s.id}", value: s.value %>
+    <%= text_field :settings, "shinycms_setting_#{s.id}", value: s.value, method: :delete %>
```
eliotsykes: Ooops, move `method: :delete` to the `link_to` below. Relevant route in `config/routes.rb` will need updating too.

PR #25, opened 2019-10-08
Page feature odds and ends

denny:
Pulling across a mixture of page-related stuff; better tests (using Faker in factories now); the start of contextually-aware main site menu; settings for default page and default section, and improvements to the code that returns both; etc blah.


app/models/page.rb
```diff
@@ -4,7 +4,11 @@ class Page < ApplicationRecord
   validates :title,       presence: true
   validates :slug,        presence: true
   validates :template_id, presence: true
-  validates :hidden,      inclusion: { in: [ true, false ] }
+  validates :hidden,      inclusion:  { in: [ true, false ] }
+  validates :slug,        uniqueness: {
```
eliotsykes: Consider adding a corresponding composite, unique index on the slug and section_id table columns, to enforce this constraint.

app/models/page.rb
```diff
@@ -37,11 +41,15 @@ def self.default_page
 
     return Page.top_level_pages.first if Page.top_level_pages.first
 
-    Page.first
+    Page.find_by( hidden: false )
```
eliotsykes: Consider adding an `order` to this query, `.find_by` could return different `Page`s on each call.

app/models/page_section.rb
```diff
@@ -3,7 +3,11 @@ class PageSection < ApplicationRecord
   validates :name,   presence: true
   validates :title,  presence: true
   validates :slug,   presence: true
-  validates :hidden, inclusion: { in: [ true, false ] }
+  validates :hidden, inclusion:  { in: [ true, false ] }
+  validates :slug,   uniqueness: {
+    scope: :section,
+    message: 'The slug must be unique within its section'
```
eliotsykes: As above, consider further enforcing uniqueness with a db index.

PR #31, opened 2019-10-10
Add IP whitelist to admin area

denny:


app/controllers/admin_controller.rb
```diff
@@ -2,6 +2,19 @@
 class AdminController < ApplicationController
   layout 'admin/layouts/admin_area'
 
+  before_action :check_admin_ip_whitelist
```
eliotsykes: This `before_action` config is inherited. The `AdminController` subclasses don't need to do this too. The `before_action :check_admin_ip_whitelist` can be removed from the subclasses `Admin::SettingsController` etc.

app/controllers/admin_controller.rb
```diff
@@ -2,6 +2,19 @@
...
+  # Check whether a whitelist of permitted admin IP addresses has been defined,
+  # and if one has, then enforce it
+  def check_admin_ip_whitelist
+    allowed = Setting.get I18n.t( 'admin_ip_whitelist' )
+    return if allowed.blank?
+
+    return if allowed.strip.split( /\s*,\s*|\s+/ ).include? request.remote_ip
+
+    redirect_to pages_path
+  end
```
eliotsykes: Consider adding test coverage for this IP address protection.

app/controllers/admin_controller.rb
```diff
@@ -2,6 +2,19 @@
 class AdminController < ApplicationController
   layout 'admin/layouts/admin_area'
 
+  before_action :check_admin_ip_whitelist
+
+  # Check whether a whitelist of permitted admin IP addresses has been defined,
+  # and if one has, then enforce it
+  def check_admin_ip_whitelist
```
eliotsykes: Instead of `whitelist`, consider using `safelist` or `allowlist`. These terms more accurately describe their purpose and have less baggage.

PR #34, opened 2019-10-11
REMOVE ALL THE THINGS !!!

denny:
Can I really comment all of these out and still have my application work?! Apparently so!

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,6 +1,22 @@
 require_relative 'boot'
 
-require 'rails/all'
+# Replace this:
+# require 'rails/all'
+# ... with this stack of individual require statements (so I can get selective)
+
+require 'rails'
+
+# require 'action_cable/engine' # Don't think I'll ever need this
+# require 'action_controller/railtie' # Thought I'd need this though?!
+# require 'action_mailbox/engine' # Won't need for a long time: inbound email
+# require 'action_mailer/railtie' # Will need soon: registration emails, etc
+# require 'action_text/engine'
+# require 'action_view/railtie'
+# require 'active_job/railtie'
+require 'active_record/railtie'
+# require 'active_storage/engine' # Will need soon: user profile pics, etc
+# require 'rails/test_unit/railtie'
+# require 'sprockets/railtie'
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
--- a/config/application.rb
+++ b/config/application.rb
@@ -6,15 +6,15 @@ require_relative 'boot'
 
 require 'rails'
 
-# require 'action_cable/engine' # Don't think I'll ever need this
-# require 'action_controller/railtie' # Thought I'd need this though?!
-# require 'action_mailbox/engine' # Won't need for a long time: inbound email
-# require 'action_mailer/railtie' # Will need soon: registration emails, etc
+# require 'action_cable/engine'     # Don't think we'll ever need this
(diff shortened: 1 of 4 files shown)


denny:
@eliotsykes This seems drastic, and I'm kind of surprised it worked. Have I crippled anything important here? :)

config/application.rb
```diff
@@ -1,6 +1,22 @@
...
 
-require 'rails/all'
+# Replace this:
+# require 'rails/all'
+# ... with this stack of individual require statements (so I can get selective)
+
+require 'rails'
+
+# require 'action_cable/engine' # Don't think I'll ever need this
+# require 'action_controller/railtie' # Thought I'd need this though?!
```
eliotsykes: I'm surprised this is working!

My guess is what's happening is that the essential ones (action_controller, action_view, maybe sprockets) are still getting loaded, but they are being `require`-d elsewhere and loading later. This probably isn't desirable as the load order may be sensitive but not creating any obvious bugs yet. I'd uncomment those essential ones here to be on the safe-side.

eliotsykes: Another thought, it may be that these still are loaded if spring is running the app in the background. Try running `bin/spring stop` then re-run app. It *might* blow up at this point. You can check spring status with `bin/spring status` too.

denny: Stopped spring, everything still seems to be working.

Being loaded further down/in was my guess, is there any way I can dig into that? I'd rather explicitly load what I have to, rather than being saved by magic/luck at a distance :)

denny: Oh, speaking of load order; I alphabetised this require list, one of my semi-OCD habits. Is that foolish, or probably safe enough?

eliotsykes: > Oh, speaking of load order; I alphabetised this require list, one of my semi-OCD habits. Is that foolish, or probably safe enough?

I'd recommend reverting it back to what Rails has. I'm not certain, but I would suspect that load order is important. Less importantly, it'll simplify reading diffs on that section of the file when you update to later Rails versions.

eliotsykes: > Being loaded further down/in was my guess, is there any way I can dig into that to know for sure what's getting loaded and where? I'd prefer to explicitly load what I have to, rather than being saved by magic/luck at a distance :)

I strongly recommend loading these here, as I do suspect there is some load order sensitive behaviour in Rails, this feels like an unnecessary gamble that could lead to tricky-to-trackdown load-order bugs.

However, I understand wanting to know what is loading these gems and where, so to find that out:

1. Comment out one of the `requires`, e.g. `#require 'action_controller/railtie'`
2. Find the commented out require file on your filesystem. `action_controller/railtie.rb` will be in a subdirectory of whatever path `bundle show actionpack` gives you.
3. At the top of that file, add some debugging like:

```
puts "-"*80
puts "In action_controller/railtie.rb, called by"
puts caller # or Thread.current.backtrace
puts "-"*80
```

This will print out a backtrace which you can follow to find out what loaded the file. Note what file loads when may differ in different environments:

- In development and test (with default config `config.eager_load = false`), some files are reloaded on each request and others are only loaded on demand, and others are only loaded once at boot by Spring.
- In production (`config.eager_load = true`), most/all files will be loaded at boot. Spring does not run.

Furthermore, the order gems are listed in `Gemfile` affects the load order of gems and their dependencies. As there are different groups of gems in `Gemfile` for each environment, this can further vary the load order per environment.

PR #39, opened 2019-10-13
A bunch of improvements to the site settings admin area...

denny:
... that totally went into the wrong branch tra la la.

### DIFF
--- a/app/controllers/admin/settings_controller.rb
+++ b/app/controllers/admin/settings_controller.rb
@@ -17,23 +17,29 @@ class Admin::SettingsController < AdminController
 
   # Add a new setting item
   def create
-    @setting = Setting.new( setting_params )
+    setting = Setting.new( setting_params )
 
-    if @setting.save
+    if setting.save
       flash[ :notice ] = I18n.t 'setting_created'
     else
       flash[ :alert ] = I18n.t 'setting_create_failed'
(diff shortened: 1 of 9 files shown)


app/controllers/admin/settings_controller.rb
```diff
@@ -17,24 +17,30 @@ def update
...
+    setting = Setting.new( setting_params )
 
-    if @setting.save
+    if setting.save
       flash[ :notice ] = I18n.t 'setting_created'
     else
       flash[ :alert ] = I18n.t 'setting_create_failed'
     end
     redirect_to admin_settings_path
+  rescue ActiveRecord::NotNullViolation, ActiveRecord::RecordNotFound
```
eliotsykes: What are the conditions that an `ActiveRecord::RecordNotFound` error happens in `create`? Similarly for the `NotNullViolation`. I'm wondering if this rescue is necessary.

denny: Not null violation is if you try to create one with no name, iirc. RecordNotFound doesn't make any sense for a create though, does it.

denny: Hrm, or maybe not. I must have got it mixed up with the update notnull - I shall remove it forthwith! :)

PR #41, opened 2019-10-15
Fix factory faker fails

denny:
Not sure what fixed it exactly, but the slug munging seems to be working now. **shrug**

### DIFF
--- a/spec/request/pages_spec.rb
+++ b/spec/request/pages_spec.rb
@@ -1,6 +1,10 @@
 require 'rails_helper'
 
 RSpec.describe 'Pages', type: :request do
+  before :all do
+    Faker::UniqueGenerator.clear
+  end
+
   describe 'GET /' do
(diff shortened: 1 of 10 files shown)


spec/request/pages_spec.rb
```diff
@@ -1,6 +1,10 @@
 require 'rails_helper'
 
 RSpec.describe 'Pages', type: :request do
+  before :all do
+    Faker::UniqueGenerator.clear
+  end
```
eliotsykes: `Faker::UniqueGenerator.clear` is repeated in 3 before hooks in 3 different specs. This does feel like state that should be cleared before every spec (much like cleaning the db, clearing the job queue, and clearing mail deliveries between tests). Consider moving it out of the specs and into a dedicated `spec/support/faker.rb` file:

```rb
# File: spec/support/faker.rb
RSpec.configure do |config|
  config.before(:each) do
    Faker::UniqueGenerator.clear
  end
end
```

PR #45, opened 2019-10-15
Batch update for admin site settings

denny:


denny:
@eliotsykes The code to only update settings with changed values feels messy; have I badly reinvented a wheel here?

PR #48, opened 2019-10-16
Move building of menu data structures into a concern, ...

denny:
… so menus can work in all areas of main site.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,6 @@
 class ApplicationController < ActionController::Base
   layout 'main_site'
 
-  include Menu
+  include MainSiteMenu
   before_action :build_menu_data
 end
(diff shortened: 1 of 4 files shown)


denny:
@eliotsykes Is this the right approach?

eliotsykes:
The approach will work OK, though it will be doing the work of building the menu object graph on every request, even if there's a response that doesn't require it, e.g. redirects, JSON responses, error responses.

One approach to avoid this would be delaying building the menu object graph as late as possible, when the menu HTML is rendered.

To do this, the method/methods would go in a helper, and they'd be called from the menu template, e.g.

```rb
# File app/helpers/menu_helper.rb
module MenuHelper
  def menu_data
    # TODO Build menu data
  end

  def another_method
    ...
  end
end
```

All methods in `MenuHelper` are available in all view templates. So to access/debug `menu_data`:

```erb
<%= menu_data.inspect %>
<%= another_method.inspect %>
```

PR #50, opened 2019-10-17
Progress on user features

denny:


app/controllers/application_controller.rb
```diff
@@ -4,4 +4,30 @@ class ApplicationController < ActionController::Base
...
   include MainSiteMenu
   before_action :build_menu_data
+
+  # Strong params config for Devise
+  before_action :configure_permitted_parameters, if: :devise_controller?
+
+  # Check logins against pwned password service and warn user if necessary
+  def after_sign_in_path_for( resource )
+    # :nocov:
+    set_flash_message! :alert, :warn_pwned if resource.respond_to?( :pwned? ) &&
```
eliotsykes: Consider reformatting to use multiline `if condition... end`.

app/controllers/application_controller.rb
```diff
@@ -4,4 +4,30 @@ class ApplicationController < ActionController::Base
...
+                                              resource.pwned?
+    super
+    # :nocov:
+  end
+
+  protected
+
+  # Strong params config for Devise
+  def configure_permitted_parameters
+    keys1 = %i[ username email password password_confirmation ]
```
eliotsykes: Consider extracting `keys*` variables in to constants.

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,14 @@
+# Controller for user features not provided by Devise - profile pages, etc
+class UserController < ApplicationController
```
eliotsykes: Consider renaming to `UsersController` - the pluralized form is more conventional and plays better with some Rails features, in particular `resource` and `resources` routes, e.g. in `config/routes.rb`:

```rb
# Singular resources
resource :user # assumes controller is named UsersController

# Plural resources
resources :users # also assumes controller is named UsersController
```

https://guides.rubyonrails.org/routing.html#singular-resources

app/models/user.rb
```diff
@@ -4,4 +4,36 @@ class User < ApplicationRecord
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :validatable,
          :confirmable, :lockable, :timeoutable, :trackable
+  devise :pwned_password unless Rails.env.test?
+
+  # Validations for username field
+  validates :username, presence:   true
+  validates :username, uniqueness: true
+  # Restrict the character set for usernames to letters, numbers, and - . _
+  validates :username, format: /\A[-_\.a-zA-Z0-9]+\z/
```
eliotsykes: Consider applying a min and max length validation to `username`.

app/models/user.rb
```diff
@@ -4,4 +4,36 @@ class User < ApplicationRecord
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :validatable,
          :confirmable, :lockable, :timeoutable, :trackable
+  devise :pwned_password unless Rails.env.test?
+
+  # Validations for username field
+  validates :username, presence:   true
+  validates :username, uniqueness: true
```
eliotsykes: Consider making username uniqueness validation case-insensitive using the `case_sensitive: false` option so variations of `denny` like `Denny` would be prevented.

PR #51, opened 2019-10-17
Tidy up path-handling code in Pages routes and controller

denny:
Also, added a top-level route that passes through to Pages controller.

### DIFF
--- a/spec/request/admin/pages_spec.rb
+++ b/spec/request/admin/pages_spec.rb
@@ -44,6 +44,21 @@ RSpec.describe 'Admin: Pages', type: :request do
       expect( response.body ).to include I18n.t 'page_create_failed'
     end
 
+    it 'fails when the page slug collides with a controller namespace' do
+      skip 'Write code to detect and prevent this'
+      template = create :page_template
+
+      post admin_page_new_path, params: {
+        'page[name]': 'Test',
+        'page[title]': 'Test',
+        'page[slug]': 'user',
+        'page[template_id]': template.id
+      }
+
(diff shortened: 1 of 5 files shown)


spec/request/admin/pages_spec.rb
```diff
@@ -44,6 +44,21 @@
       expect( response.body ).to include I18n.t 'page_create_failed'
     end
 
+    it 'fails when the page slug collides with a controller namespace' do
```
eliotsykes: A custom validator using `Rails.application.routes.recognize_path(path, method:)` will allow us to check against existing routes. Here's a skeleton of how that might work:

```rb
# File: app/validators/routes_validator.rb
#
# Usage in model:
#   validates :slug, routes: { to: 'pages#show' }
#
# Expects model to have a `#path` method that returns the URL path (excluding leading slash) this
# model will be located at.
class RoutesValidator < ActiveModel::EachValidator

  def validate_each(record, attribute, value)
    return if value.blank? && options[:allow_blank]

    record.errors.add(attribute, 'does not route correctly') unless routes_ok?(record.path)
  end

  private

  def routes_ok?(path)
    route_params = Rails.application.routes.recognize_path(path, method: :get)
    "#{route_params[:controller]}##{route_params[:action]}" == options[:to] && path == route_params[:path]
  end

end
```

This validator expects the model to have a `#path` method, e.g. for `Page`:

```rb
def path
  # TODO: return the full path (omit leading slash), including parent
  # page slugs, e.g. "recipes/desserts/apple-crumble"
  slug
end
```

PR #53, opened 2019-10-18
Add elements to page templates

denny:


app/controllers/admin/pages/templates_controller.rb
```diff
@@ -42,9 +41,40 @@ def update
 
   private
 
+  # Create the template elements, based on the contents of the template file
+  def create_elements
```
eliotsykes: For `create_elements` and `update_elements` consider:

- performing their work inside a single transaction that'll rollback if any individual `save`/`update` fails, to avoid incomplete data being committed
- moving this work into methods in the model most responsible for this work

denny: I think I found a Rails magic to do it all instead...?

https://github.com/denny/ShinyCMS-ruby/pull/54/files - specifically https://github.com/denny/ShinyCMS-ruby/pull/54/files#diff-7e513a48b9eb3ca69ee6cf5636beb663R14

eliotsykes: Ace 👍

PR #54, opened 2019-10-19
Template elements saving in update...

denny:
... using appropriate levels of Rails magic :-p

Also, renamed element attributes because 'type' is a reserved word.

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -2,7 +2,8 @@
 class UserController < ApplicationController
   def index
     if user_signed_in?
-      render :show, params: { username: current_user.username }
+      @user_profile = User.find_by( username: current_user.username )
+      render :show
     else
       redirect_to user_login_path
     end
--- a/app/models/page_template.rb
+++ b/app/models/page_template.rb
@@ -5,11 +5,13 @@ class PageTemplate < ApplicationRecord
 
   has_many :pages, foreign_key: 'template_id',
                    inverse_of: 'template',
-                   dependent: :delete_all
+                   dependent: :nullify
   has_many :elements, class_name: 'PageTemplateElement',
                       foreign_key: 'template_id',
                       inverse_of: 'template',
                       dependent: :delete_all
+
+  accepts_nested_attributes_for :elements
 
   # Instance methods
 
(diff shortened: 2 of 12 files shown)


app/models/page_template.rb
```diff
@@ -5,12 +5,14 @@ class PageTemplate < ApplicationRecord
...
   has_many :pages, foreign_key: 'template_id',
                    inverse_of: 'template',
-                   dependent: :delete_all
+                   dependent: :nullify
   has_many :elements, class_name: 'PageTemplateElement',
                       foreign_key: 'template_id',
                       inverse_of: 'template',
                       dependent: :delete_all
 
+  accepts_nested_attributes_for :elements
```
eliotsykes: 👍 for `accepts_nested_attributes_for` usage

app/controllers/user_controller.rb
```diff
@@ -2,7 +2,8 @@
 class UserController < ApplicationController
   def index
     if user_signed_in?
-      render :show, params: { username: current_user.username }
+      @user_profile = User.find_by( username: current_user.username )
```
eliotsykes: Can the `find_by` be avoided?

```diff
-@user_profile = User.find_by( username: current_user.username )
+@user_profile = current_user
```

denny: Oh wow. How embarrassing is that! Coding when tired :rofl:

app/models/page_template.rb
denny: I've just written up [the blog post for this PR](http://shinycms.org/blog/2019/10/magic) and a year later, I'm still pleased with myself for actively seeking out and finding (some sort of magic in the region of) `accepts_nested_attributes_for` (and getting it working), exactly when I needed it, all on my own :stuck_out_tongue: This was the first time I felt like a good native Rails developer rather than a barely competent non-idiomatic one :slightly_smiling_face: 

This was when the project proved itself as a learning tool for me (a lot earlier than I had any right to expect or even hope, really!) - also, ironically, this is when I got interested enough and confident enough that it stopped being just a personal educational project and turned into a thing I wanted to see through to a release. Which of course has made it all the more educational. :smiley:

PR #55, opened 2019-10-19
Allow username logins as well as email logins

denny:


app/controllers/application_controller.rb
```diff
@@ -21,13 +21,17 @@ def after_sign_in_path_for( resource )
...
-    devise_parameter_sanitizer.permit :sign_up, keys: keys1
+    basic_params = %i[ username email password password_confirmation ]
+    devise_parameter_sanitizer.permit( :sign_up ) do |user_params|
+      user_params.permit( basic_params )
+    end
 
-    keys2 = %i[ login username email password password_confirmation remember_me]
-    devise_parameter_sanitizer.permit :sign_in, keys: keys2
+    devise_parameter_sanitizer.permit( :sign_in ) do |user_params|
+      user_params.permit( basic_params | %i[ login remember_me ] )
```
eliotsykes: Consider using `#merge` as this is more commonly used in ruby and may be clearer than `|` for this operation:

```diff
-basic_params | %i[ login remember_me ]
+basic_params.merge %i[ login remember_me ]
```

denny: I think I was running up against the cops for Style/MethodLength and Style/LineLength here. Is there any community consensus on non-default settings for those, or on when it's okay to override them?

eliotsykes: I don't think there's a consensus, but the 80 chars limit is considered outdated by some devs (not just Ruby devs) given current screen resolutions. Some options:

- Up the MaxLineLength to 120 (or 100 if 120 it too much to stomach). (I use 120 in my editor as line guide, but I don't enforce it with linting/rubocop.)
- Extract those arrays into constants. This will reduce the number of objects allocated in memory when this method is called which is a small bonus.

eliotsykes: Also I got this wrong:

> Consider using #merge as this is more commonly used in ruby and may be clearer than | for this operation:

At some point I wrongly thought this method was dealing in hashes. Its clearly using arrays. `#merge` won't work here with `Array`s. Ignore my original advice from yesterday - sorry!

PR #56, opened 2019-10-19
Helpers and Concerns and Elements, oh my!

denny:
Move controller concern for building menus into a helper instead.
Create a model concern for common features across all the various *Element types (current and future!)
Tweak menu partial to handle scruffy menu data a bit better (although still plenty to do)


app/views/includes/_menu.html.erb
```diff
@@ -1,34 +1,48 @@
-<% unless @menu_tl_sections.blank? %>
+<% menu_data = build_menu_data( @page&.section ) %>
 <ul>
+<% unless menu_data[ :pages ][ :top_sections ].blank? %>
```
eliotsykes: Nested hashes like `menu_data` signal a candidate for a new class, such as `Menu`. Consider refactoring this. Here's a sketch of one approach:

```rb
# File: app/models/menu.rb
class Menu
  
  attr_accessor :top_pages, :top_sections, :section

  def initialize(section = nil)
    self.section = section if section
    self.top_sections = PageSection.top_level_sections
    self.top_pages = Page.top_level_pages
  end

  def sections
    section&.sections || []
  end

  def pages
    section&.pages || []
  end

end
```

```rb
# In template or in helper
menu = @page.menu

# In Page model
def menu
  Menu.new(section)
end

# In menu partial
<% if menu.top_sections.any? %>
  ...
<% end %>

<% menu.top_sections.each do |top_section| %>
  ...
<% end %>
```

Over time, I'd expect this new class to attract more menu-related behaviour, gathering those responsibilities together, and improving readability and maintainability.

PR #60, opened 2019-10-23
Page creation now builds element stack (based on page template)

denny:


app/models/concerns/name_title_slug.rb
```diff
@@ -0,0 +1,30 @@
...
+    validates :slug, safe_slug: true
+    validates :slug, format:    ANCHORED_SLUG_REGEX
+    validates :slug, uniqueness: {
+      scope: :section,
+      message: I18n.t( 'slug_unique' )
+    }
+  end
+
+  class_methods do
+    def slugify( slug )
```
eliotsykes: Consider using `parameterize` in or instead of `slugify`: https://api.rubyonrails.org/classes/String.html#method-i-parameterize

config/application.rb
```diff
@@ -33,6 +33,9 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
+    # Add app/validators to autoload paths
+    config.autoload_paths << Rails.root.join( 'app', 'validators' )
```
eliotsykes: Subdirectories of `app/` shouldn't need to be added to `config.autoload_paths`, line 37 can be removed.

> Autoload paths automatically pick any custom directories under app. For example, if your application has app/presenters, or app/services, etc., they are added to autoload paths. https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoload-paths

denny: Did not work on my dev environment (worked fine in tests and CI mind you). Have I broken my local setup somehow?

app/models/concerns/name_title_slug.rb
```diff
@@ -0,0 +1,30 @@
+# Common behaviour for names, titles, and slugs, for pages and page sections
+module NameTitleSlug
```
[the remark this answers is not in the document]
denny: Hopefully the name seems a bit more appropriate after today's additions? :)

config/application.rb
eliotsykes: It might have been that any rails servers and spring processes needed to be restarted after the `app/validators/` directory was added. Perhaps try these steps in order:

1. removing the `config.autoload_paths << ...` line
2. stopping any running rails servers
3. Run `bin/spring stop` to stop any spring processes
4. start a rails server `bin/rails s`
5. check the safe slug validator still works

app/models/concerns/name_title_slug.rb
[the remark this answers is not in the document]
eliotsykes: Agreed, it does look better after the additions.

PR #61, opened 2019-10-23
Enable/disable user login feature via site settings

denny:


app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
+# Controller for user features not provided by Devise - profile pages, etc
+class Users::SessionsController < Devise::SessionsController
+  def new
+    unless Setting.yes? I18n.t( 'allow_user_logins' )
```
eliotsykes: Consider improving the readability of this line. The combination of the `yes?` and the `I18n.t(...)` feels like it could be clarified with a change like:

```diff
-unless Setting.yes? I18n.t( 'allow_user_logins' )
+unless Setting.allow_user_logins?
```

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
+# Controller for user features not provided by Devise - profile pages, etc
+class Users::SessionsController < Devise::SessionsController
+  def new
```
eliotsykes: It looks like other `Devise::SessionsController` actions (e.g. `create`) should prevent logins too. If yes, consider using a `before_action` to protect multiple actions/methods in one go.

app/models/setting.rb
```diff
@@ -2,10 +2,28 @@
...
 
+  SOUNDS_LIKE_YES = %w[ yes y true on ].freeze
+  private_constant :SOUNDS_LIKE_YES
+
+  # Instance methods
+
+  # Return true if a setting has a yes-ish value
+  def yes?
+    return true if value.downcase == I18n.t( 'yes' ).downcase
+    return true if SOUNDS_LIKE_YES.include? value.downcase
```
eliotsykes: Consider moving the normalization to a `before_validate` or `before_save` callback so the db stores truthy/yessy values consistently as a single value, and the normalization only needs to happen once at create/update time, and not on reads.

PR #62, opened 2019-10-24
Generate page/section slugs/titles if they aren't specified by user

denny:


app/views/pages/templates/example.html.erb
```diff
@@ -6,11 +6,11 @@
...
-  <%= @page.description %>
+  <%= @page.element( 'para1' ) %>
 </p>
 <p>
-  <img src="<%#= @page.elements.image_name %>">
+  <img src="<%= @page.element( 'image_name' ) %>">
 </p>
 <p>
-  <%#= @page.elements.para2 %>
+  <%= @page.element( 'para2' ) %>
```
eliotsykes: Hi @denny - I see you got `<%= page.element('para1') %>` working.

You were interested in getting `<%= page.elements.para1 %>` working. One approach that *might* work is using association extensions: https://guides.rubyonrails.org/association_basics.html#association-extensions, and implementing `method_missing` and `respond_to?` on the module. I expect the `method` arg would be `:para1` in the above example.

```rb
# File: app/models/concerns/element_reader.rb
module ElementReader
  def method_missing(method)
    # For page.elements.para1, method would be :para1
    where(name: method).pick(:content) || super
  end

  def respond_to?(method, include_private_methods = false)
    if super
      true
    else
      exists?(name: method)
    end
  end
end 

# File app/models/page.rb
class Page < ApplicationRecord
  ...

  has_many :elements, -> { extending ElementReader }
    class_name: 'PageElement',
    foreign_key: 'page_id',
    inverse_of: 'page',
    dependent: :delete_all

```

---

In case you wanted something even terser, we could find ways to make any of these work (these wouldn't use association extensions):

```erb
<%= @para1 %>

<%= para1 %>

<%= element :para1 %>

<%= page_element :para1 %>
```


---

One thing to consider about `erb` templates. They are unsafe in that they can run any code, so a malicious template can do a lot of damage, access the db, leak a lot of data, etc., etc.

To protect against this, consider using a safer, restricted templating language such as handlebars for the custom user-provided templates. You can still use erb for other templates within the app.

denny: Well I had a lot of fun failing to get this to work last night :D I'll push it up to a branch and link you, hopefully you can tell me where I went wrong! :)

denny: @eliotsykes Booyah. https://github.com/denny/ShinyCMS-ruby/commit/dc87c3566da2032a0c42ef8da09eed42b2286bb5

denny: That said, I just looked at how ShinyCMS-perl deals with elements and it's not quite what I'd remembered (which is embarrassing!) and that lead me onto a thought; if I pre-build a hash of the page elements available when I load the page, then dynamically turning that into a list of reader methods for that page would be a lot simpler than supporting any arbitrary method on any page, right? I might take a swing at that approach tonight.

denny: Took a swing at this approach using define_method and couldn't get that to work either; I look forward to discussing my failures with you at some point :)

eliotsykes: > dc87c35

@denny I'll need your help debugging `ElementReader` next call.

> if I pre-build a hash of the page elements available when I load the page, then dynamically turning that into a list of reader methods for that page would be a lot simpler than supporting any arbitrary method on any page, right? I might take a swing at that approach tonight.

It might be possible to get something like this working using the `method_missing` technique in a helper (instead of pre-defining the methods), e.g. `PageHelper` in `app/helpers/page_helper.rb`. Sketched out implementation (warning: likely buggy!) below:

```rb
# File app/helpers/page_helper.rb
module PageHelper
  # To support <%= element1 %> when @page.elements.exists?(name: 'element1')
  def method_missing( method, *args )
    if @page && content = @page.elements.where(name: method).pick(:content)
      return content
    end
    super
  end
end
```

eliotsykes: Mmm...perhaps there's a simpler solution that doesn't involve metaprogramming with `method_missing`. This feels less likely to cause subtle bugs with naming collisions.

In `PagesController#show_page`, we can pass a `locals` hash into the `render template: ...` call. These `locals` will be available in the template:

```diff
def show_page
    ...

-    render template: "pages/templates/#{@page.template.filename}"
+    render template: "pages/templates/#{@page.template.filename}",
+      locals: { element1: ..., some_html: ..., etc. }
  end
```

The `locals:` object would need to be built from `@page.elements`.

Then in erb `<%= element1 %>`, `<%= some_html %>` would be supported.

PR #67, opened 2019-11-05
Shared content admin area

denny:
Adds an admin area page for creating and updating shared content.

### DIFF
--- a/app/controllers/admin/shared_content_controller.rb
+++ b/app/controllers/admin/shared_content_controller.rb
@@ -10,9 +10,10 @@ class Admin::SharedContentController < AdminController
     element = SharedElement.new( new_element_params )
 
     if element.save
-      flash[ :notice ] = I18n.t 'admin.shared_content.element_created'
+      flash[ :notice ] = I18n.t 'admin.shared_content.shared_content_created'
     else
-      flash[ :alert ] = I18n.t 'admin.shared_content.element_create_failed'
+      flash[ :alert ] =
+        I18n.t 'admin.shared_content.shared_content_create_failed'
     end
     redirect_to admin_shared_content_path
   end
@@ -22,21 +23,21 @@ class Admin::SharedContentController < AdminController
     updated_something = false
     updated_something = update_shared_content( updated_something )
     flash[ :notice ] = if updated_something
-                         I18n.t 'admin.shared_content.elements_updated'
+                         I18n.t 'admin.shared_content.shared_content_updated'
                        else
-                         I18n.t 'admin.shared_content.elements_unchanged'
+                         I18n.t 'admin.shared_content.shared_content_unchanged'
                        end
     redirect_to admin_shared_content_path
   end
 
   # Delete an existing shared content element
   def delete
-    if SharedContent.destroy( params[ :id ] )
-      flash[ :notice ] = I18n.t 'admin.shared_content.element_deleted'
+    if SharedElement.destroy( params[ :id ] )
+      flash[ :notice ] = I18n.t 'admin.shared_content.shared_content_deleted'
     end
     redirect_to admin_shared_content_path
   rescue ActiveRecord::NotNullViolation, ActiveRecord::RecordNotFound
-    flash[ :alert ] = I18n.t 'admin.shared_content.element_delete_failed'
+    flash[ :alert ] = I18n.t 'admin.shared_content.shared_content_delete_failed'
     redirect_to admin_shared_content_path
   end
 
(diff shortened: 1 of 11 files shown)


app/views/admin/shared_content/index.html.erb
```diff
@@ -0,0 +1,36 @@
+<% @page_title = t( 'admin.shared_content.shared_content' ) %>
+
+<%= form_with name: 'all_shared_content', url: admin_shared_content_path, method: :post do %>
+<table class="list">
+  <% @elements.each do |e| %>
+  <tr class="showfocus">
+    <td>
+      <%= text_field :shared_content, "element_#{e.id}_name", value: e.name %>
```
eliotsykes: I suspect there is a way to get the `<input>`s named in such a way that the resulting `params` is a more usable data structure and simplify `SharedContentController#update_shared_content`.

Sadly I'm not certain how to do this off the top of my head and can't find a good reference, but we could go through this together if you're interested. I'd approach it by:

- Tinkering with using `fields_for` https://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-fields_for to see how that goes and see what kind of name attributes it generates for the `<input>`s.
- Review these on using param names that get parsed into arrays of hashes:
  - https://stackoverflow.com/questions/3089849/ruby-on-rails-submitting-an-array-in-a-form#3090026
  - https://medium.com/@dmccoy/how-to-build-html-inputs-to-submit-an-array-of-hashes-with-a-rails-form-tag-that-work-with-strong-f70a6c03d78e

app/views/admin/shared_content/index.html.erb
```diff
@@ -0,0 +1,36 @@
...
+		</td>
+  </tr>
+  <% end %>
+</table>
+<p>
+  <%= submit_tag t( 'admin.update' ) %>
+</p>
+<% end  # form %>
+
+<%= form_with name: 'new_element', url: admin_shared_content_create_path, method: :post do %>
```
eliotsykes: Check if this form is friendly and remembers last entered values when the user submits invalid data. If it doesn't and it should, consider:

- providing a `model: @shared_element` option to the `form_with` call
- using the FormBuilder object (often named `f`) that `form_with` yields to its block `form_with ... do |f| ... end`. Use `f` to build the fields, e.g. `f.text_field ...`.
- modifying the relevant controller actions to support this.

app/controllers/admin/shared_content_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    elements = shared_content_params[ :shared_content ]
+    elements&.each_key do |key|
+      next unless /element_(?<id>\d+)_content/ =~ key
+
+      content = elements[ "element_#{id}_content" ]
+      content_type = elements[ "element_#{id}_content_type" ]
+
+      element = SharedElement.find( id )
+      next if element.content == content && element.content_type == content_type
+
```
eliotsykes: To see if a model has changes, consider making use `ActiveModel::Dirty`'s methods: https://api.rubyonrails.org/classes/ActiveModel/Dirty.html - `changed?` in particular looks relevant here.

app/views/admin/shared_content/index.html.erb
denny: Let's discuss this tomorrow; I've been wondering if I could tidy this up a lot by having a model that is to SharedElements what a Page is to PageElements - then I could use fields_for and make all the update code simpler, just saving the parent object. What I don't know if whether I'd need to create a table and model with one SharedContent 'page' in it to achieve this, or whether I could do something a bit more abstract to tie it all together without cluttering the schema - I suspect the latter will be feasible and probably more elegant.

The same situation applies to Settings btw, and there it occurs to me that a SettingsCollection object would be useful for implementing user-overridable settings at some point down the line.

app/controllers/admin/shared_content_controller.rb
denny: I was going to ask whether there was something like this I should be using. Ta!

app/views/admin/shared_content/index.html.erb
eliotsykes: >  I could do something a bit more abstract to tie it all together without cluttering the schema with a single row table. I feel like the latter is probably very doable, but I'm not sure what terms to google to get started on approaching it!

Perfect, this sounds like a job for form objects. Rough sketch to kick us off later:

```
# File: app/forms/foo_form.rb
class FooForm
  # ActiveModel::Model provides validations and other model niceties:
  include ActiveModel::Model

  attr_accessor :some_collection, :some_attribute
  ...
end
```

In the controller:

```rb
@foo_form = FooForm.new
```

And then pass `@foo_form` to `form_for/form_with`.

PR #68, opened 2019-11-06
Rename shared content elements

denny:
Groundwork for some refactoring (coming soon! watch this space! your investment may go down as well as sideways!)


eliotsykes:
`SharedContentElement` certainly works as a name, I'm just wondering if there's something more succinct and less than 3 words. Perhaps `Snippet`, `Fragment` or ?

PR #71, opened 2019-11-07
Add CKEditor for HTML page elements

denny:


config/routes.rb
```diff
@@ -30,6 +30,9 @@
     # ========== ( Admin area ) ==========
     get 'admin', to: 'admin#index'
 
+    # CKEditor (WYSIWYG editor used on various admin pages)
+    mount Ckeditor::Engine => '/admin/ckeditor'
```
eliotsykes: Are the ckeditor routes only accessible to admins?

PR #76, opened 2019-11-20
Add CoreUI templates and styles to admin area

denny:
Also, fix CKEditor config issue.


app/views/admin/includes/_breadcrumbs.html.erb
```diff
@@ -1,11 +1,10 @@
 <nav id="breadcrumbs">
-  <%= link_to t( 'admin.home' ),  url_for( '/' ) %>
-  &raquo;
-  <%= link_to t( 'admin.admin' ), url_for( admin_path ) %>
-  &raquo;
+<ol class="breadcrumb border-0 m-0">
+  <li class="breadcrumb-item"><%= link_to t( 'admin.home' ),  url_for( '/' ) %></li>
```
eliotsykes: - `url_for(admin_path)` also available as `admin_url`
- `url_for( '/' )` also available as `root_url`

app/views/admin/includes/_head.html.erb
```diff
@@ -1,16 +1,19 @@
...
+    <%= favicon_link_tag %>
 
-<%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload' %>
-<%= javascript_pack_tag 'admin_area',  'data-turbolinks-track': 'reload' %>
+    <%= stylesheet_link_tag 'https://unpkg.com/@coreui/coreui@3.0.0-beta.1/dist/css/coreui.min.css' %>
+    <%= stylesheet_link_tag 'https://fonts.googleapis.com/css?family=Poppins&display=swap'          %>
+    <%= stylesheet_link_tag 'shinycms/admin_area', media: 'all', 'data-turbolinks-track': 'reload'  %>
 
-<%= javascript_include_tag Ckeditor.cdn_url %>
+    <% if action_name == 'edit' || controller_name == 'shared_content' %>
```
eliotsykes: For clarity, consider extracting a method that reveals the intent of this condition, while also hiding the implementation details behind it, e.g.

```erb
<%= javascript_include_tag Ckeditor.cdn_url if editable? %>
```

```rb
def editable?
  action_name == 'edit' || ...
end
```

The extracted method would go in one of the `app/helpers/` files.

app/views/admin/includes/_messages.html.erb
```diff
@@ -1,10 +1,57 @@
+<% if flash[ :notice ] %>
```
eliotsykes: Consider removing the duplication from this template.

app/views/themes/halcyonic/layouts/halcyonic.html.erb
```diff
@@ -1,11 +1,11 @@
-<!DOCTYPE HTML>
+<!DOCTYPE html>
 <!--
 	This set of templates is heavily based on Halcyonic by HTML5 UP:
 	https://html5up.net/halcyonic | html5up.net | @ajlkn
 	Free for personal and commercial use under the CCA 3.0 license:
   html5up.net/license
 -->
-<% if controller_path == 'pages' && @page == Page.default_page %>
+<% if controller_name == 'pages' && @page == Page.default_page %>
```
eliotsykes: Consider replacing the `@viewing_homepage` instance variable with a helper method in one of `app/helpers/*.rb`, e.g.

```rb
def viewing_homepage?
  controller_name == 'pages' && @page == Page.default_page
end
```

In template:

```erb
<% if viewing_homepage? %>
  ...
<% end %>
```

app/views/admin/includes/_head.html.erb
denny: Ooh, that's satisfyingly neat :)

app/views/themes/halcyonic/layouts/halcyonic.html.erb
denny: Rubocop yelled at me for using the @page instance variable in a helper, so I tidied it up with a method in the Pages model instead... what do you think? https://github.com/denny/ShinyCMS-ruby/commit/e1e907512887395fea0e09927e94c701798dd221

app/views/admin/includes/_messages.html.erb
denny: Like so? https://github.com/denny/ShinyCMS-ruby/commit/ed35c1d2957860f2a4537e6e4304cd23ea133c08

eliotsykes: Good start, there's a tad more possible https://github.com/denny/ShinyCMS-ruby/pull/78#discussion_r349530168

app/views/themes/halcyonic/layouts/halcyonic.html.erb
eliotsykes: Yep looks good 👍 

I don't think the following is needed here but it can be handy in similar situations (it'd also satisfy Rubocop). 

View helper methods can also be defined in controllers, using `helper_method`: https://api.rubyonrails.org/classes/AbstractController/Helpers/ClassMethods.html#method-i-helper_method

Here, we could define a `current_page` method on `PagesController` and make it available to the view:

```rb
class PagesController ...

  # Makes methods available to views
  helper_method :current_page, :viewing_homepage?

  def current_page
    @page
  end

  def viewing_homepage?
    current_page.default_page?
  end
```

PR #78, opened 2019-11-22
Assorted minor fixes in admin area

denny:
Added a helper method for whether to load CKEditor, tidied up messages partial, fixed a couple of odd/inconsistent uses of path/url helper methods.

### DIFF
--- a/app/helpers/elements_helper.rb
+++ b/app/helpers/elements_helper.rb
@@ -13,4 +13,9 @@ module ElementsHelper
     filenames = Dir.glob '*?.?*', base: path
     filenames.sort
   end
+
+  # Return true if the page we're on might need a WYSIWYG HTML editor
+  def html_editor_needed?
+    return true if action_name == 'edit' || controller_name == 'shared_content'
+  end
 end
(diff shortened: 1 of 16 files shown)


app/helpers/elements_helper.rb
```diff
@@ -13,4 +13,9 @@ def select_filenames( path = nil )
     filenames = Dir.glob '*?.?*', base: path
     filenames.sort
   end
+
+  # Return true if the page we're on might need a WYSIWYG HTML editor
+  def html_editor_needed?
+    return true if action_name == 'edit' || controller_name == 'shared_content'
```
eliotsykes: Line can be shortened to just the `if` condition.

app/views/admin/includes/_messages.html.erb
```diff
@@ -1,57 +1,13 @@
...
 
-<% if flash[ :alert ] %>
-<div class="alert alert-danger alert-dismissible fade show" role="alert">
-  <%= flash[ :alert ] %>
-  <button type="button" class="close" data-dismiss="alert" aria-label="Close">
-    <span aria-hidden="true">&times;</span>
-  </button>
-</div>
-<% end %>
+<%= render partial: 'admin/includes/message', locals: { message_type: 'danger',  message: flash[ :alert_danger  ] } if flash[ :alert_danger  ] %>
```
eliotsykes: Consider drying a little further:

```rb
[:danger, :warning, :success, :info, :light, :dark].each do |message_type|
  <% next unless message = flash[:"alert_#{message_type}"] %>
  <%= render partial: '...', locals: { message_type: message_type, message: message } %>
end
```

PR #79, opened 2019-11-22
Minor fixes for main site

denny:
In particular, make alerts visible. Also, a helper method for figuring out when we're looking at the homepage.

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -43,6 +43,11 @@ class Page < ApplicationRecord
     hash
   end
 
+  # Return true if this page is the default page
+  def default_page?
+    return true if self == Page.default_page
+  end
+
   # Class methods
 
   def self.top_level_pages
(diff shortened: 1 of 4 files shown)


app/models/page.rb
```diff
@@ -43,6 +43,11 @@ def elements_hash
     hash
   end
 
+  # Return true if this page is the default page
+  def default_page?
+    return true if self == Page.default_page
```
eliotsykes: Line 48 can be shortened to `self == Page.default_page`

denny: Is there a downside to keeping it more explicit?

eliotsykes: There's not much of a downside to keeping it as is, if any. The very minor points in its favour are:

- `return true if self == Page.default_page` returns `true` or `nil` (never `false`). The method return type is not consistently boolean, compared to below.
- `self == Page.default_page` returns `true` or `false`. The method return type is consistently boolean, which is a little more predictable but admittedly nothing great.
- the shorter version is more conventional.

PR #82, opened 2019-11-22
Move menu state js into separate file

denny:
Still need to get this (and ckeditor config) to play nice with webpacker at some point.

### DIFF
--- a/app/views/admin/menu/_menu.html.erb
+++ b/app/views/admin/menu/_menu.html.erb
@@ -22,3 +22,6 @@
   <%= render partial: 'admin/menu/menu_item', locals:
       { text: 'Settings', link: admin_settings_path, icon: 'settings' } %>
 </ul>
+<script>
+  persist_menu_state();
+</script>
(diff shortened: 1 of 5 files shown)


app/views/admin/menu/_menu.html.erb
```diff
@@ -22,3 +22,6 @@
   <%= render partial: 'admin/menu/menu_item', locals:
       { text: 'Settings', link: admin_settings_path, icon: 'settings' } %>
 </ul>
+<script>
+  persist_menu_state();
```
eliotsykes: Should `persist_menu_state()`'s work be delayed until after the DOM has finished loading? 

It looks like the function searches the document  immediately with `document.getElementsByClassName( 'c-sidebar-nav-dropdown' )` which could be called before the DOM finishes loading and so miss some elements.

denny: Probably, yeah. I'm not very happy about anything in the recent JavaScript commits :expressionless:  So long as it works for me then I'm inclined to ignore it for now, and revisit it when I have time to learn more and redo it properly - or wait until somebody else downloads the project and fixes it or files an issue :smiley:

PR #85, opened 2019-11-27
Add user admin section (and paging). Split user profiles off from Devise.

denny:


app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      flash[ :notice ] = I18n.t 'admin.users.user_updated'
+      redirect_to action: :edit, id: @user.id
+    else
+      flash.now[ :alert ] = I18n.t 'admin.users.user_update_failed'
+      render action: :edit
+    end
+  end
+
+  def delete
+    if User.destroy( params[ :id ] )
```
eliotsykes: Is it possible for a logged-in user to accidentally delete themselves here?

app/views/admin/menu/_menu_pages.html.erb
```diff
@@ -0,0 +1,13 @@
+  <%= render partial: 'admin/menu/menu__section_start', locals:
+      { text: 'Pages', icon: 'list-rich' } %>
+      <%= render partial: 'admin/menu/menu__item', locals:
```
eliotsykes: Consider introducing a helper method to make these repeated calls a little more succinct, e.g.

```diff
-<%= render partial: 'admin/menu/menu__item', locals: { text: 'List Pages', link: ..., icon: '...' } %>
+<%= render_menu_item 'List pages', link: ... icon: '...' %>
```

app/models/user.rb
```diff
@@ -19,16 +19,15 @@ class User < ApplicationRecord
...
+  # Configure default count-per-page for pagination
+  paginates_per 20
+
   # Virtual attribute to allow authenticating by either username or email
   attr_writer :login
   def login
     @login || username || email
   end
 
-  def to_param
```
eliotsykes: Regarding this commit: 

> Unless I can have to_param work differently in main site and admin area, it needs to go to id 
https://github.com/denny/ShinyCMS-ruby/pull/85/commits/74a4c594bbb7d9e30e82f5a59548c7e7ff27da2e

Consider making use of `CurrentAttributes` to make this possible using request-thread-scoped variables. These docs are well worth a read for an overview: https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html

A sketch of how this could work:

```rb
class User ...
  
  def to_param
    if Current.area == :admin
      id
    else
      username
    end
  end
```

In `app/controllers/application_controller.rb`, set the current area to something like public/main/non-admin in a before action:

```rb
class ApplicationController ...
  before_action :set_current_area
  
  def set_current_area
    Current.area = :public
  end
```

In `app/controllers/admin_controller.rb`, override `set_current_area`.

```rb
class AdminController ...

  def set_current_area
    Current.area = :admin
  end
```

Create `app/models/current.rb` to make this all work:

```rb
class Current < ActiveSupport::CurrentAttributes
  attribute :area
end
```

PR #86, opened 2019-12-01
Working Heroku deploy

denny:
* rake db:migrate in release phase
* fight with asset pipeline (just for a change)
* fix some image tags
* add action_cable back in (not sure why this became necessary here)


config/application.rb
```diff
@@ -6,7 +6,7 @@
 
 require 'rails'
 
-# require 'action_cable/engine'     # Don't think we'll ever need this
+require 'action_cable/engine'       # Why do I need this?
```
eliotsykes: I don't know why its needed, that's unexpected. 

My advice would be to reorder these `require`s so they are in the same order that `rails/all.rb` performs them, as I suspect the wrong load order could cause subtle bugs.

After that, what happens if action_cable is commented out? If there's an error, whats the message + stack trace?

PR #87, opened 2019-12-01
Add 'if not exists' flag to migrations

denny:
Also added 'show in menu' option to pages and page sections.

### DIFF
--- a/db/migrate/20190928091301_create_page_sections.rb
+++ b/db/migrate/20190928091301_create_page_sections.rb
@@ -1,7 +1,7 @@
 # Create table for page sections
 class CreatePageSections < ActiveRecord::Migration[6.0]
   def change
-    create_table :page_sections do |t|
+    create_table :page_sections, if_not_exists: true do |t|
       t.string :name, null: false
       t.text :description
       t.string :title, null: false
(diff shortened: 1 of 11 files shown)


db/migrate/20190928091301_create_page_sections.rb
```diff
@@ -1,16 +1,16 @@
...
-    create_table :page_sections do |t|
+    create_table :page_sections, if_not_exists: true do |t|
       t.string :name, null: false
       t.text :description
       t.string :title, null: false
       t.string :slug, null: false
       t.integer :default_page_id
       t.integer :section_id
       t.integer :sort_order
+      t.boolean :hidden_from_menu, null: false, default: false
```
eliotsykes: Consider writing new migrations to modify existing tables.

denny: As discussed, the reason I keep editing old 'CreateTable' migrations instead of adding new 'AddColumn' ones is the lack of the AFTER keyword in Postgres, and my distaste for the disordered tables that result.

That said, I was thinking that dropping the whole table and recreating it with the new column would be a bad thing because data loss (!) but what if I renamed the current table, created the new one, and then copied the data across - are there Rails-y constructs for doing it that way?

eliotsykes: Here are the methods that would help with doing this: https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html

Methods that look particularly relevant include `columns(table_name)`, `indexes(table_name)`, `rename_table`, `drop_table`, `rename_column`, `remove_column`.

Rather than recreating the table and indexes, it may be slightly less work doing the technique described here: https://wiki.postgresql.org/wiki/Alter_column_position#Add_columns_and_move_data

PR #89, opened 2019-12-01
More work on feature flags

denny:


config/locales/en.yml
```diff
@@ -12,40 +12,30 @@ en:
...
-
-    features:
-      users:
-        can_register: Allow users to register
-        can_login: Allow users to login
-
 
   # ========== ( Main Site ) ==========
 
+  features:
```
eliotsykes: Rails i18n supports variable interpolation which allows for:

```yaml
features:
  alert: "Sorry, the %{feature_name} feature of this site has been turned off"
```

```rb
I18n.translate('features.alert', feature_name: 'Foo')
#> "Sorry, the Foo feature of this site has been turned off"
```

Search for mentions of "interpolat" in https://guides.rubyonrails.org/i18n.html for more details.

db/migrate/20191127225110_create_feature_flags.rb
```diff
@@ -0,0 +1,11 @@
+class CreateFeatureFlags < ActiveRecord::Migration[6.0]
+  def change
+    create_table :feature_flags do |t|
+      t.string :name
```
eliotsykes: Consider adding an index to the `name` column as its likely this column is going to be queried frequently. Guessing it should also be a unique index to prevent duplicates.

app/helpers/features_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+    return false if feature.blank?
+    return false if feature.off?
+    return true  if feature.on?
+
+    return false unless current_user&.admin?
+
+    feature.admin_only?
+  end
+
+  def feature_alert_message( feature_name )
```
eliotsykes: This can likely be replaced using Rails i18n variable interpolation support (see other comment).

spec/rails_helper.rb
```diff
@@ -69,4 +69,7 @@
   # Allow request specs to use have_* matchers, like so:
   # expect( response.body ).to have_title 'My Page Title'
   config.include Capybara::RSpecMatchers, type: :request
+
+  # Load my app's helpers
+  config.include FeaturesHelper, type: :request
```
eliotsykes: Try to avoid including app/ helpers in specs, feels risky, it might lead to some unexpected method overrides. Perhaps this can be removed if i18n variable interpolation is used?

To test a helper, this can be done in `spec/helpers/`, in a dedicated spec.

spec/spec_helper.rb
```diff
@@ -8,6 +8,10 @@
   SimpleCov.formatter = SimpleCov::Formatter::Codecov
 end
 
+# Make sure tests go through the 'bad theme' code in config/application.rb
+ENV['SHINYCMS_THEME'] = 'TEST'
+Dir.mkdir 'app/views/themes/TEST' unless Dir.exist? 'app/views/themes/TEST'
```
eliotsykes: Should this directory be cleaned up after the tests run? If so, consider configuring a RSpec `around` hook (or `before`+`after` hooks)  to create and delete the directory.

PR #93, opened 2019-12-04
Bump versions of ruby and various gems, node and various modules

denny:


eliotsykes:
As dependencies are up-to-date since this PR, it seems like a good time to set up dependabot.

denny:
> As dependencies are up-to-date since this PR, it seems like a good time to set up dependabot.

Ah! Thank you!! I kept going over the list of CI (etc) services in my head because I had this feeling that I'd definitely forgotten something that I have on other projects, but no matter how many times I went over it in my head I could not figure out what it was. Probably because dependabot doesn't have a cute badge to add to my README file :stuck_out_tongue: 

Done now :)

eliotsykes:
Nicely done!

What do you think about adding [Brakeman](https://github.com/presidentbeef/brakeman) security scans to this project? It could be setup as a step of the CI build or using https://docs.codeclimate.com/docs/brakeman.

PR #95, opened 2019-12-04
Prevent loading of Action Cable in production

eliotsykes:
@denny I think these are all the references to Action Cable. 

My *guess* (I've not tried to see if this fixes it) is the ruby references (in particular app/channels/) are responsible for loading action cable in production. When the production environment boots, it will eagerly load all .rb files under app/.

To find all these references, I grepped (well "`ag`ged" really) all of the application's gems for `skip_action_cable`:

```sh
$ ag -C 10 -u skip_action_cable
```

`skip_action_cable` is the setting used during `rails new my-new-app` to determine if action cable's artifacts should be created in the new app or not.


denny:
Nice one, thanks! I did a trawl at the weekend and found most of these, but not quite all of them :)

PR #108, opened 2019-12-15
Refactor capabilities and Pundit auth

denny:
Also, fixed auth problem in front of 'bad delete' tests... which means they're back in... which means we're back to 100% coverage :tada:


app/controllers/admin/pages/templates_controller.rb
```diff
@@ -53,7 +54,8 @@ def delete
     flash[ :notice ] = t( 'template_deleted' ) if template.destroy
     redirect_to admin_pages_templates_path
   rescue ActiveRecord::NotNullViolation, ActiveRecord::RecordNotFound
```
eliotsykes: Consider removing the duplicated `rescue ...` lines throughout the `delete` controller actions by passing a block, e.g.

```rb
def delete
  handle_delete_exceptions(t('template_delete_failed'), admin_pages_templates_path) do
    template = PageTemplate.find( params[:id] )
    authorise template
    flash[ :notice ] = t( 'template_deleted' ) if template.destroy
    redirect_to admin_pages_templates_path
  end
end
```

```rb
def handle_delete_exceptions(alert_message, redirect_path)
  yield
rescue ActiveRecord::NotNullViolation, ActiveRecord::RecordNotFound
  skip_authorization
  redirect_to redirect_path, alert: alert_message
end
```

db/demo-site-data/users.yml
```diff
@@ -0,0 +1,26 @@
+---
+User_1:
+  id: 1
+  username: admin
+  email: admin@example.com
+  encrypted_password: "$2a$11$BuIQa84a0DBc9vAuJ.PubOb3K7L8.UlMcOPltxRGU9j2FfIQkNolC"
```
eliotsykes: This is public. I wanted to highlight in case this hash protects a sensitive value that should not be exposed, even hashed.

eliotsykes: Although this is for a demo site, consider generating unique random values for sensitive data (password, confirmation_token + ?) to mitigate risk of demo sites being compromised by a reused password/token. 

Perhaps there's a possibility the demo admin user is not deleted/reset when a demo site is converted to a working site?

denny: At some point I want to add a curl-able install.sh script, so I'll have that prompt people to choose their own username and password for the admin account.

PR #127, opened 2019-12-22
Fix Faker uniqueness

denny:
I kept getting test failures from element names like 'tin' and similar colliding with parts of words on my pages (in expect(...).not_to include(...) tests), which was tedious. So, Lorem; it's not as much fun as Science, but it's less likely to contain parts of real words. Meh.


denny:
Hilariously, this failed on its first commit, due to using 'ut' as a word. FFS.

eliotsykes:
Could some of these collisions be avoided by targeting the expectations more specifically, say by using the `have_css` instead of `include` matchers?

PR #134, opened 2019-12-26
Blogs (WIP reached working milestone, so merging fwiw)

denny:
(AKA the "I got 99 commits but a blog feature ain't one oh wait yes it is" PR)


app/models/blog_post.rb
```diff
@@ -0,0 +1,32 @@
+# Model class for blog posts
+class BlogPost < ApplicationRecord
+  belongs_to :user
```
eliotsykes: Consider renaming so it explains the association's intent:

```diff
-belongs_to :user
+belongs_to :author, class_name: 'User'
```

PR #136, opened 2019-12-28
Refactor: resourceful(ish) routes

denny:
Alter routing config to use the built-in Rails resource method (albeit with a lot of special casing, to get rid of the inappropriate and wildly irritating (yes I know it's just me) pluralisation in paths for singular entities).


denny:
PS: This would have been almost impossible to get right without 100% test coverage; w00t.

denny:
@eliotsykes It's a Xmas miracle ;)

PR #137, opened 2019-12-29
i18n and more granular capability checks in admin menus/toolbar

denny:


app/views/admin/menu/_menu_other.html.erb
```diff
@@ -0,0 +1,9 @@
+  <%= render_admin_menu_section_start( t( 'admin.other.title' ), 'settings' ) %>
```
eliotsykes: `render_admin_menu_section_start` and `render_admin_menu_section_end` are candidates for combining into a single method that accepts a block arg:

```erb
<%= render_admin_menu_section(t( 'admin.other.title' ), 'settings') do %>
  ...
<% end %>
```


```rb
def render_admin_menu_section(text, icon=nil)
  render_admin_menu_section_start(text, icon)
  yield
  render_admin_menu_section_end
end
```

This makes it so forgetting to call `render_admin_menu_section_end` at the end of these templates is not possible, making it less likely invalid HTML will be rendered.

denny: @eliotsykes Just tried this and for some reason I'm losing the start section HTML when it renders the menu... I verified that it's getting into the _start method and reaching the end of it (warn statement each side of the render call), and the log says it rendered the template in-between those two, but the actual output HTML shows no trace of that being true!

End section template is rendering, just to make it even weirder. Is it something to do with how the text and icon params are passed around maybe?

Branch here: https://github.com/denny/ShinyCMS-ruby/tree/refactor/menu-sections

eliotsykes: See feedback at https://github.com/denny/ShinyCMS-ruby/commit/04c5e5ec6f88c007634a5736c17fea392dd61bb4#r36842232

PR #138, opened 2019-12-29
Tidy up and test month/year view of blog posts

denny:


app/models/blog.rb
```diff
@@ -29,25 +29,40 @@ class Blog < ApplicationRecord
...
+
+  def posts_for_year( year )
+    start_date = "#{year}-01-01".to_date
+    end_date = start_date.clone + 1.year
+    posts.where(
+      'posted_at between ? and ?',
+      start_date.to_s,
+      end_date.to_s
+    ).order( :posted_at )
+  end
```
eliotsykes: ActiveRecord supports using `Range`s for dates and will generate a SQL `BETWEEN` query: https://guides.rubyonrails.org/active_record_querying.html#range-conditions

Consider using this in the 2 methods above, e.g.

```rb
start_date = Date.new(year, 1, 1)
# clone not needed
end_date = start_date + 1.year
posts.where(posted_at: start_date..end_date).order(:posted_at)
```

PR #150, opened 2020-01-11
Smart Menus

denny:
Incorporate jQuery Smart Menus plugin for main site menus/submenus, and add a menu item for blog with submenu for recent blog posts.


app/views/themes/shinycms/menu/_blog.html.erb
```diff
@@ -0,0 +1,13 @@
+<% Blog.all.each do |blog| %>
+<li>
+  <%= link_to blog.name, view_blog_path( blog ) %>
+  <ul>
+    <% menu_posts = blog.recent_posts %>
```
eliotsykes: Consider adding a `limit(5)` to the `blog.recent_posts` call to retrieve the only those posts that are needed from the db.

PR #157, opened 2020-01-16
FeatureFlag.update_all_flags returns validation result

eliotsykes:


app/controllers/admin/feature_flags_controller.rb
```diff
@@ -9,9 +9,8 @@ def index
 
   def update
     authorise FeatureFlag
-    @flags = FeatureFlag.update_all_flags( feature_flag_params )
 
-    if @flags.all?( &:valid? )
+    if FeatureFlag.update_all_flags( feature_flag_params )
```
eliotsykes: @denny Try stubbing `FeatureFlag.update_all_flags` in the spec to test when this fails, e.g.:

```rb
expect(FeatureFlag).to receive(:update_all_flags).and_return(false)
```

app/models/feature_flag.rb
```diff
@@ -4,16 +4,20 @@ class FeatureFlag < ApplicationRecord
 
   # Class methods
 
+  # Returns true if all valid, else false
   def self.update_all_flags( params )
+    all_valid = true
     params['flags'].keys.map do |flag_id|
       flag = find( flag_id )
       flag_params = params['flags'][flag_id]
-      _f = flag.update(
```
eliotsykes: Rubocop was trying to tell us something!

denny: Hah! Convergent evolution!  https://github.com/denny/ShinyCMS-ruby/commit/f5e490347d522c65ae3c4a8c7816fc3ec84448f4

PR #167, opened 2020-01-28
Improve post-login redirect

denny:
... and add login form to bottom of Halcyonic theme pages

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -45,7 +45,9 @@ class ApplicationController < ActionController::Base
   def check_for_pwnage( resource )
     return unless resource.respond_to?( :pwned? ) && resource.pwned?
 
+    # :nocov:
     set_flash_message! :alert, :warn_pwned
+    # :nocov:
   end
 
   # Override post-login redirect
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -54,8 +54,8 @@ class ApplicationController < ActionController::Base
   def after_sign_in_path_for( resource )
     check_for_pwnage( resource )
 
-    if request.referer.present?
-      return request.referer unless request.referer == new_user_session_url
+    if request.referer.present? && request.referer != new_user_session_url
+      return request.referer
     end
 
     return admin_path if resource.can? :view_admin_area
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -41,17 +41,23 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  # Check user's password against pwned password service and warn if necessary
+  def check_for_pwnage( resource )
+    return unless resource.respond_to?( :pwned? ) && resource.pwned?
+
+    set_flash_message! :alert, :warn_pwned
+  end
+
+  # Override post-login redirect
   def after_sign_in_path_for( resource )
-    # Check user's password against pwned password service and warn if necessary
-    if resource.respond_to?( :pwned? ) && resource.pwned?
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -40,15 +40,25 @@ def configure_permitted_parameters
     end
   end
 
+  # Check user's password against pwned password service and warn if necessary
+  def check_for_pwnage( resource )
+    return unless resource.respond_to?( :pwned? ) && resource.pwned?
```
eliotsykes: This is very minor, the line can be slightly simplified using `#try`:

```diff
-return unless resource.respond_to?( :pwned? ) && resource.pwned?
+return unless resource.try( :pwned? )
```

eliotsykes: I can't remember how the `pwned?` check works in the test environment. 

It could be making a HTTP request to a remote API every time a test logs a user in, slowing the tests down or possibly breaking them if the haveibeenpwned API is down. 

Remote API calls are usually best avoided during tests for these reasons.

We can workaround this a few ways I expect. Let me know if its making remote API calls.

app/controllers/application_controller.rb
```diff
@@ -40,15 +40,25 @@ def configure_permitted_parameters
...
   def after_sign_in_path_for( resource )
-    # Check user's password against pwned password service and warn if necessary
-    if resource.respond_to?( :pwned? ) && resource.pwned?
-      # :nocov:
-      set_flash_message! :alert, :warn_pwned
-      # :nocov:
+    check_for_pwnage( resource )
+
+    if request.referer.present? && request.referer != new_user_session_url
+      return request.referer
```
eliotsykes: I'm not sure, but there may be a potential security vulnerability here, if the referer is a 3rd-party (attacker-owned) site.

Perhaps if combined with other vulnerabilities, ShinyCMS could be made to redirect a user to a phishing copy site pretending to be the genuine site.

Consider only returning the referer path here, removing any host. 

~~For extra confidence, consider checking the referer path against the app's known `GET` routes using `Rails.application.routes.recognize_path`.~~ <- Safelist probably overkill now I think about it, but returning only the path without the host is worthwhile.

spec/requests/admin/users_spec.rb
```diff
@@ -125,9 +125,9 @@
         expect( response      ).to     have_http_status :ok
         expect( response.body ).to     have_title I18n.t( 'admin.users.index.title' ).titlecase
         expect( response.body ).to     have_css '.alert-success', text: I18n.t( 'admin.users.destroy.success' )
-        expect( response.body ).to     include u1.username
-        expect( response.body ).not_to include u2.username
-        expect( response.body ).to     include u3.username
+        expect( response.body ).to     have_css 'td', text: u1.username
+        expect( response.body ).not_to have_css 'td', text: u2.username
+        expect( response.body ).to     have_css 'td', text: u3.username
```
eliotsykes: If it makes the intent clearer and the test more readable, consider using Capybara's `have_table` matcher along with one of its `cols`,`rows`,`with_cols`,`with_rows` options to replace lines 128-130. 

- https://rubydoc.info/github/teamcapybara/capybara/master/Capybara%2FRSpecMatchers:have_table
- Options documented at https://rubydoc.info/github/teamcapybara/capybara/master/Capybara%2FNode%2FMatchers:has_table%3F

app/controllers/application_controller.rb
denny: https://github.com/denny/ShinyCMS-ruby/blob/master/app/models/user.rb#L8 skips this during tests.

PR #168, opened 2020-01-28
Fix month names in blog month/year views

denny:
... and re-enable tests that check for them.

### DIFF
--- a/spec/requests/blogs_spec.rb
+++ b/spec/requests/blogs_spec.rb
@@ -87,7 +87,7 @@ RSpec.describe 'Blogs', type: :request do
       get "/blog/#{post1.posted_year}/#{post1.posted_month}"
 
       expect( response      ).to     have_http_status :ok
-      # TODO: expect( response.body ).to include 'February'
+      expect( response.body ).to     include 'February'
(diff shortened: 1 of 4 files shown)


spec/requests/blogs_spec.rb
```diff
@@ -87,7 +87,7 @@
       get "/blog/#{post1.posted_year}/#{post1.posted_month}"
 
       expect( response      ).to     have_http_status :ok
-      # TODO: expect( response.body ).to include 'February'
+      expect( response.body ).to     include 'February'
```
eliotsykes: Consider using the `have_css` matcher over the `include` matcher to more precisely target the expectation, and less prone to false positives.

spec/requests/blogs_spec.rb
```diff
@@ -104,8 +104,8 @@
       get "/blog/#{post1.posted_year}"
 
       expect( response      ).to have_http_status :ok
-      # TODO: expect( response.body ).to include 'February'
-      # TODO: expect( response.body ).to include 'September'
+      expect( response.body ).to include 'February'
+      expect( response.body ).to include 'September'
```
eliotsykes: As above, favour `have_css` over `include` matcher.

PR #170, opened 2020-01-30
Add tags to blog posts, and TagsController to view all tags on site

denny:


app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    @tagged_items = {}
+    @content_types.each do |type|
+      @tagged_items[ type ] = type.constantize.tagged_with( @tag.name )
+    end
+  end
+
+  private
+
+  def tagged_content_types
+    %w[ BlogPost ]
```
eliotsykes: Browsing the acts_as_taggable source, it looks like it adds the class method `taggable?` to all models, so if you wanted to avoid having to keep this array up to date, consider something like:

```
def tagged_content_types
  # Select all model classes that return SomeModel.taggable? as true:
  ApplicationRecord.descendants.select(&:taggable?)
end
```

app/views/shinycms/tags/show.html.erb
```diff
@@ -0,0 +1,17 @@
+<% @content_types.each do |type| %>
+  <h2>
+    <%= type.underscore.pluralize.humanize %> tagged with '<%= @tag %>'
+  </h2>
+  <ul>
+    <% @tagged_items[ type ].each do |item| %>
+    <li>
+      <% if type == 'BlogPost' %>
```
eliotsykes: (Eventually) This might be a job for `polymorphic_path` to save the type checking code growing here when generating the URL: https://api.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html

```diff
-<%= link_to item.title, view_blog_post_path( item ) %>
+<%= link_to item.title, polymorphic_path( item ) %>
```

PR #176, opened 2020-01-31
Minor fixes

denny:
Make default 404 more reusable. Prevent non-existent usernames crashing the profile page.


app/controllers/users_controller.rb
```diff
@@ -12,6 +12,10 @@ def index
 
   def show
     @user_profile = User.find_by( username: params[ :username ] )
```
eliotsykes: Consider if its more appropriate to show a 404 rather than redirect to the home page when there's a non existent username in the path.

If you want this 404 behaviour, raise an `ActiveRecord::RecordNotFound` error, which Rails will handle by rendering the default not found page with HTTP status 404. To raise this error, call `find_by!` instead of `find_by`:

```diff
-@user_profile = User.find_by( username: params[ :username ] )
+@user_profile = User.find_by!( username: params[ :username ] )
```

If you opt for this, lines 15-18 can be removed.

PR #177, opened 2020-02-01
Recursive page/section menus

denny:


app/models/page_section.rb
```diff
@@ -46,6 +46,22 @@ def default_page
...
   end
 
+  def menu_pages
+    pages.where( hidden_from_menu: false )
+  end
+
+  def menu_sections
+    sections.where( hidden_from_menu: false )
+  end
+
```
denny: @eliotsykes I seem to have opted for implementing menu_pages/menu_sections as methods (just above), but hidden_pages/hidden_sections as relationships (starting on line 13). Are there any reasons to prefer one or the other? I guess I should pick one and stick to it, at least :)

eliotsykes: @denny Mmm, I don't think I have a strong argument for this.

My feeling is the association-heavy approach could lead to the number of associations becoming unwieldy to work with, compared to the alternative.

The approach I suggest would be to reduce the number of associations, so you have one association with *all* pages, and one assocation with *all* sections:

```rb
has_many :pages, -> { order(:sort_order) }, dependent: :restrict_with_error
has_many :sections, -> { order(:sort_order) }, 
            class_name: 'PageSection',
            foreign_key: 'section_id',
            inverse_of: 'section',
            dependent: :restrict_with_error
```

Then introduce instance methods to handle the variations for `#hidden_pages`, `#visible_pages`, `#hidden_sections`, `#visible_sections`, as you've done for `#menu_sections` and `#menu_pages`.

If nothing else, this approach reduces the repeated assocation config for `class_name`, `foreign_key` and so on.

PR #181, opened 2020-02-03
Fix user capability admin

denny:


app/views/admin/users/_capability_checkbox.html.erb
```diff
@@ -1,3 +1,4 @@
+<%= 'SKIP' if capability == 'skip' %>
```
eliotsykes: Consider replacing the if/unless on lines 1 & 2 with an if/else.

denny: Huh. That line was debug output ... that never worked and hence didn't get removed! :D

PR #182, opened 2020-02-07
Refactor settings: stage 1

denny:
Refactor of settings stuff to handle storing/serving/updating the site-wide/default settings, now with room for users/admins to override selected settings (post-login redirect, I'm looking at you).

### DIFF
--- a/app/controllers/admin/site_settings_controller.rb
+++ b/app/controllers/admin/site_settings_controller.rb
@@ -1,12 +1,12 @@
-# Admin controller for CMS settings
-class Admin::SettingsController < AdminController
+# Admin controller for site settings
+class Admin::SiteSettingsController < AdminController
   def index
     @settings = Setting.order( :name )
-    authorise @settings
+    authorise @settings.first
   end
 
   def update
-    skip_authorization # TODO: this is obviously bad; refactoring v. soon!
+    authorise Setting.all.first
 
     all_updated = update_settings( settings_params )
     if all_updated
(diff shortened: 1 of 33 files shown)


app/controllers/admin/site_settings_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    else
+      flash[ :alert ] = t( '.failure' )
+    end
+    redirect_to site_settings_path
+  end
+
+  private
+
+  def update_settings( params )
+    flag = true
```
eliotsykes: Consider renaming the `flag` variable to reveal its intent.

app/controllers/application_controller.rb
```diff
@@ -68,6 +68,6 @@ def set_view_paths
     # Add the default templates directory to the top of view_paths
     prepend_view_path 'app/views/shinycms'
     # Apply the configured theme, if any, by adding it above the defaults
-    prepend_view_path Theme.current.view_path if Theme.current
+    prepend_view_path Theme.current.view_path if Theme.current( current_user )
```
eliotsykes: Should the `.current` call in `Theme.current.view_path` pass in the `current_user` arg too?

spec/models/setting_spec.rb
```diff
@@ -3,8 +3,35 @@
...
+      user = create :admin_user
+
+      expect( Setting.get( :foo, user ) ).to eq 'bar'
     end
   end
+
+  it 'fails validation when updating a locked setting' do
+    setting = create :setting, name: 'foo', level: 'site', locked: true
+
+    _null = setting.update( level: 'user' )
```
eliotsykes: Consider adding an expectation on this line too, testing `#update` returns false when `setting` is invalid:

```diff
-_null = setting.update( level: 'user' )
+expect(setting.update( level: 'user' )).to eq false
```

PR #185, opened 2020-02-08
Main site page for user-configurable site settings

denny:


app/controllers/admin/site_settings_controller.rb
```diff
@@ -9,40 +9,71 @@ def update
     authorise Setting.all.first
 
     all_updated = update_settings( settings_params )
-    if all_updated
+    if all_updated.nil?
```
eliotsykes: Having `all_updated == false` and `all_updated == nil` mean two different things feels a bit risky.

Perhaps `#update_settings` could return a result that more clearly defines its possible values, such as a symbol: `:unchanged`, `:success`, `:failure`.

app/controllers/admin/site_settings_controller.rb
```diff
@@ -9,40 +9,71 @@ def update
...
+    flags[ :no_change ] = true
+    flags[ :no_errors ] = true
+    flags
+  end
 
-    val.update!( value: value )
+  def update_flags( flag, flags )
+    flags[ :no_change ] = false unless flag.nil?
+    flags[ :no_errors ] = false if flag == false
+    flags
```
eliotsykes: I'm finding some of the methods in this controller a little tricky to follow. 

However I admit I've not got a better suggestion.

It *may* be that using the `ActiveModel::Dirty` methods like `#changed?` can help simplify the code here, but no guarantees: https://api.rubyonrails.org/classes/ActiveModel/Dirty.html

app/models/setting.rb
```diff
@@ -48,6 +48,14 @@ def self.get( name, user = nil )
     setting.site_value
   end
 
+  def self.user_settings
+    where( level: 'user' ).order( :name )
+  end
+
+  def self.admin_settings
+    where( level: 'user' ).or( where( level: 'admin' ) ).order( :name )
```
eliotsykes: This `.or` can be removed and the line reduced as follows:

```diff
-where( level: 'user' ).or( where( level: 'admin' ) ).order( :name )
+where( level: ['user', 'admin'] ).order( :name )
```

app/controllers/site_settings_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      flag  = update_value( id, params )
+      flags = update_flags( flag, flags )
+    end
+    return if flags[ :no_change ]
+
+    flags[ :no_errors ]
+  end
+
+  def update_value( id, params )
+    setting = Setting.find( id )
```
eliotsykes: Consider adding a further mitigation to prevent unauthorised updates, by scoping the `.find` call here:

```diff
-setting = Setting.find(id)
+settings = current_user_is_admin? Setting.admin_settings : Setting.user_settings
+setting = settings.find(id)
```

This scoping is applied in a few places, so consider extracting it into the `SettingPolicy` and making use of pundit's `policy_scope` method.

app/controllers/site_settings_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    return false if setting.level == 'site'
+    return false if setting.level == 'admin' && current_user_is_not_admin?
```
eliotsykes: If this is a blatant security violation, consider raising a custom error immediately to fail loudly and make it likely it'd get picked up in the logs and/or an exception tracker like Honeybadger, and hopefully support would be alerted to and investigate suspect behaviour from a user.

PR #191, opened 2020-02-10
Docs

denny:
Added separate doc for info about themes, fleshed out a few other bits here and there, changed some filenames to make some stand out more and some stand out less.


docs/Getting-Started.md
```diff
@@ -0,0 +1,38 @@
+# Getting Started
+
+## Deployment
+
+You can deploy ShinyCMS to Heroku using the included Procfile, after setting
```
eliotsykes: Perhaps a candidate for a Deploy To Heroku button? https://devcenter.heroku.com/articles/heroku-button

PR #194, opened 2020-02-20
Comments

denny:
Adding comments feature. Comments are hooked to Discussions, which have a polymorphic connection to a parent resource (currently just blog posts, but the future beckons). :)


app/controllers/admin/discussions_controller.rb
```diff
@@ -0,0 +1,69 @@
+# Admin controller for discussions and comments
+class Admin::DiscussionsController < AdminController
```
eliotsykes: In Rails, when you see a controller that is not using the conventional RESTful controller actions (`index`, `new`, `create`, etc.), its **sometimes** (perhaps not here) a sign to introduce a controller that would fit the conventional actions.

Here for example, we've got actions `hide`, `unhide`, `lock`, `unlock`, and `*_comment` variations of those too.

Consider if this is a sign to use or introduce `controller#actions` such as:

For discussions:

- `Admin::DiscussionsController#update` handles all hiding/unhiding, locking/unlocking
OR
- `Admin::DiscussionLocksController` and `Admin::DiscussionHidersController`:
    - `DiscussionLocksController#create` -> locks discussion
    - `DiscussionLocksController#destroy` -> unlocks discussion
    - `DiscussionHidersController#create` -> hides discussion
    - `DiscussionHidersController#destroy` -> unhides discussion

For comments:
- `Admin::CommentsController`
    - `#update` handles all hiding/unhiding, locking/unlocking
    - `#destroy` handles destroying
OR
- `Admin::CommentLocksController`, `Admin::CommentHidersController`, and `Admin::CommentsController#destroy`.

I'm not convinced these are appropriate here, but something to consider.

By going with the RESTful routes convention, the controllers tend to have more consistency, and other  advantages often appear throughout the app's development. In particular, it can lead you to a model you didn't know was needed that corresponds to the controller (e.g. `DiscussionLock`, `CommentHider`, etc. - prob. not appropriate here admittedly).

app/controllers/admin/discussions_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+  before_action :stash_discussion
+  before_action :stash_comment, except: %i[ hide unhide lock unlock ]
+
+  def hide
+    authorise @discussion
+    @discussion.hide
+    redirect_to request.referer || discussion_path( @discussion )
```
eliotsykes: Consider using `redirect_back fallback_location: discussion_path(@discussion)` https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_back

app/controllers/admin/discussions_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  private
+
+  def stash_discussion
+    @discussion = Discussion.find( params[ :id ] )
+  end
+
+  def stash_comment
+    @comment = @discussion.comments.find_by( number: params[ :number ] )
```
eliotsykes: If the comment doesn't exist, should a 404 be shown? If so, use `.find_by!`, it'll raise a `RecordNotFound` error which will be translated to a 404.

app/controllers/discussions_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+class DiscussionsController < ApplicationController
+  before_action :check_feature_flags
+
+  def index
+    # TODO: list of recently active discussions etc
+    redirect_to root_path
+  end
+
+  def show
+    @discussion = Discussion.find( params[ :id ] )
```
eliotsykes: Could this allow a user to view a hidden discussion? If yes, consider scoping to published discussions. This might be a good spot to use pundit's `policy_scope`.

app/controllers/discussions_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def show_thread
+    @discussion = Discussion.find( params[ :id ] )
+    @comment = @discussion.comments.find_by( number: params[ :number ] )
```
eliotsykes: Might this allow hidden comments and/or discussions to be seen?

app/models/comment.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  validates :discussion_id, presence: true
+  validates :number, uniqueness: { scope: :discussion_id }
+
+  validates :body,  presence: true, unless: -> { title.present? }
+  validates :title, presence: true, unless: -> { body.present?  }
+
+  # Instance methods
+
+  def lock
```
eliotsykes: There *might* be an existing `lock` instance method provided by ActiveRecord. There is a `lock!` one for sure, so this naming is a little risky.

app/models/blog_post.rb
```diff
@@ -2,8 +2,13 @@
...
   belongs_to :author, class_name: 'User',
-                      foreign_key: 'user_id',
-                      inverse_of: 'blog_posts'
+                      foreign_key: :user_id,
+                      inverse_of: :blog_posts
+
+  has_one :discussion, as: :resource, dependent: :destroy
+
+  delegate :discussion_hidden, to: :discussion, allow_nil: true
+  delegate :discussion_locked, to: :discussion, allow_nil: true
```
eliotsykes: Consider using:

```
delegate :hidden, :locked, to: :discussion, prefix: true, allow_nil: true
```

The `prefix` option means the generated methods will still be named `discussion_hidden` and `discussion_locked`, and those methods in `Discussion` can be deleted.

If you prefer the `predicate?` method naming:

```
delegate :hidden?, :locked?, to: :discussion, prefix: true, allow_nil: true
```

would generate `BlogPost#discussion_hidden?` and `#discussion_locked?` methods.

app/models/discussion.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  # Instance methods
+
+  def discussion_hidden
+    hidden
+  end
+
+  def discussion_locked
+    locked
+  end
```
eliotsykes: See comment above, `discussion_hidden` and `discussion_locked` can be deleted if `prefix` option used with `delegate`.

app/models/discussion.rb
```diff
@@ -0,0 +1,40 @@
+# Model class for discussions (used to group comments)
+class Discussion < ApplicationRecord
+  belongs_to :resource, inverse_of: :discussion, polymorphic: true
```
eliotsykes: I'm getting good feelings from this polymorphic association 😄

app/policies/admin/comment_policy.rb
```diff
@@ -0,0 +1,22 @@
+# Pundit policy for administration of comments
+class Admin::CommentPolicy < Admin::DefaultPolicy
+  def hide_comment?
```
eliotsykes: The `_comment` suffixes feel a little unnecessary - consider removing them from the `CommentPolicy` and specifying the method name in the controller `authorize` calls.

app/views/shinycms/discussions/_comment_thread.html.erb
```diff
@@ -0,0 +1,14 @@
+<% if comment.hidden %>
```
eliotsykes: ActiveRecord generates `predicate?` methods for boolean columns like `hidden`, so line 1 could read (if you prefer): `if comment.hidden?`.

db/migrate/20200218155902_add_comments.rb
```diff
@@ -0,0 +1,25 @@
+class AddComments < ActiveRecord::Migration[6.0]
+  def change
+    create_table :comments, if_not_exists: true do |t|
+      t.integer :discussion_id, null: false
+      t.integer :number, null: false
```
eliotsykes: Consider adding a composite unique index on `discussion_id` and `number`, corresponding to the uniqueness validation, and use the db index to protect the data integrity.

app/views/shinycms/discussions/_comment_posted_by.html.erb
```diff
@@ -0,0 +1,14 @@
+Posted by
+<% if comment.author_type == 'user' %>
+  <%= user_profile_link( comment.author ) %>
+<% elsif comment.author_type == 'pseudonymous' %>
+  <% if comment.author_url.present? %>
+    <%= link_to comment.author_name, comment.author_url %>
```
eliotsykes: Consider adding `rel='nofollow'` or `rel='ugc'` to author links to reduce incentive for spammers: https://support.google.com/webmasters/answer/96569?hl=en

app/views/admin/includes/_admin_toolbar.html.erb
```diff
@@ -64,8 +64,28 @@
     <%= link_to t( 'list' ), users_path %>
   <% end %>
     |
+<% end %>
 
+<% if @discussion.present? %>
+  <%= t( 'admin.discussion.title' ) %>:
+<% if current_user_can? :hide, :discussions %>
+  <% if @discussion.hidden? %>
+    <%= link_to t( 'unhide' ), unhide_discussion_path( @discussion ) %>
```
eliotsykes: Try to avoid generating `HTTP GET` links/anchor tags that make state changes. They're susceptible to link prefetching wreaking havoc (admittedly unlikely). Prefer using `button_to` which generates a button wrapped with a form. Also prefer to use an HTTP method that conveys its mutating a resource such as PATCH.

This applies to all the hide/unhide and lock/unlock links for discussions and comments in this PR.

PR #197, opened 2020-02-22
Add reCAPTCHA to registration process

denny:


app/helpers/recaptcha_helper.rb
```diff
@@ -0,0 +1,46 @@
+# Helper methods for turning CMS features on/off selectively
```
eliotsykes: This comment looks like it came from another helper.

app/helpers/recaptcha_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+module RecaptchaHelper
+  include MainSiteHelper
```
eliotsykes: Is this `include MainSiteHelper` needed? I suspect it'll work without it.

app/helpers/recaptcha_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    # TODO: fix the tests so they can get here (might need to refactor?)
+    # :nocov:
+    verify_recaptcha_v2
+    # :nocov:
+  end
+
+  def verify_invisible_recaptcha( action )
+    # Used in spec/requests/user_spec.rb, to test fallback to checkbox
+    return false if ENV[ 'FAIL_RECAPTCHA' ].present?
```
eliotsykes: Does this ENV var only exist for testing? If so, we may be able to replace it with mocking or stubbing one of these helper methods in the test.

PR #199, opened 2020-02-23
News section

denny:
Essentially a clone of the blog feature, with the sketchy multi-blog support ripped out.


app/views/shinycms/news/month.html.erb
```diff
@@ -0,0 +1,27 @@
+<% @page_title = "Blog posts in #{I18n.t("date.month_names")[@posts.first.posted_at.month]} #{@posts.first.posted_at.year}"  %>
```
eliotsykes: In a few of these news templates, the word "blog" appears a few times. Perhaps some of these are intended to be "news" instead of "blog".

PR #203, opened 2020-02-26
Move admin creation from seed data to rake task

denny:


lib/tasks/admin.rake
```diff
@@ -1,48 +1,62 @@
...
-      admin = User.new( username: username, password: password, email: email)
+# ShinyCMS admin user administration tasks
+
+# rails shiny:admin:create - creates a new superadmin user, with full privs
+
+namespace :shiny do
+  namespace :admin do
+    desc 'Create a super-admin user, with the full set of capabilities'
+    task :create, %i[ u p e ] => [ :environment ] do |_t, args|
+      username    = args[ :u ]
```
eliotsykes: To add test coverage for rake tasks, there are a couple of approaches:

**1)** Move the rake task code into a method and unit test that method. E.g. this could be moved into a `User.create_super_admin` method and tested in `spec/models/user_spec.rb`. The rake task would be reduced to something like:

```rb
task :create, %i[ u p e ] => [ :environment ] do |_t, args|
  User.create_super_admin(args)
end
```

or **2)** Introduce rake task specs as described here:
- https://www.eliotsykes.com/test-rails-rake-tasks-with-rspec
- https://github.com/eliotsykes/rails-testing-toolbox/blob/master/tasks.rb (prefer this file, rather than the one in the body of the article above, it has been improved since)

PR #210, opened 2020-02-28
Add reCAPTCHA protection to comment forms

denny:


app/controllers/blogs_controller.rb
```diff
@@ -47,6 +48,11 @@ def set_blog
     redirect_to root_path
   end
 
+  def stash_recaptcha_keys
```
eliotsykes: `stash_recaptcha_keys` appears to be duplicated in a few controllers. When there's duplicate code in a controller, there's a few options for DRYing it up, including moving it to a model, move it to a helper, and move it to the parent class `ApplicationController`.

This responsibility aligns well with `RecaptchaHelper` so consider refactoring this to move this there. Something like:

```rb
module RecaptchaHelper

  # Named public but maybe its private?
  def recaptcha_public_key_v2
    ENV['RECAPTCHA_V2_SITE_KEY']
  end

  def recaptcha_public_key_v3
    ENV['RECAPTCHA_V3_SITE_KEY']
  end
end
```

Those methods are then available to all view templates.

spec/requests/discussions_spec.rb
```diff
@@ -17,6 +17,20 @@
...
+    ENV['RECAPTCHA_V3_SECRET_KEY'] = 'XYZ'
+    ENV['RECAPTCHA_V3_SITE_KEY'  ] = 'ZYX'
+  end
+
+  after :each do
+    ENV.delete( 'RECAPTCHA_V3_SECRET_KEY' ) if @v3_secret.nil?
+    ENV.delete( 'RECAPTCHA_V3_SITE_KEY'   ) if @v3_site.nil?
+
+    ENV['RECAPTCHA_V3_SECRET_KEY'] = @v3_secret unless @v3_site.nil?
+    ENV['RECAPTCHA_V3_SITE_KEY'  ] = @v3_site   unless @v3_site.nil?
```
eliotsykes: Consider hiding the ENV vars behind methods that are then stubbed in the specs. This will mean the tests don't need to do this ENV var cleanup.

PR #216, opened 2020-03-08
Web stats (powered by Ahoy)

denny:
Lots of improvements to be made in the code for displaying/filtering/ordering the stats, but the basics are all in place now.

Merge early, merge often!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,7 +16,7 @@ class ApplicationController < ActionController::Base
 
   layout 'layouts/main_site'
 
-  after_action :track_action
+  after_action :track_ahoy_visit
 
   protected
 
@@ -76,7 +76,7 @@ class ApplicationController < ActionController::Base
   end
 
   # Track all actions with Ahoy
-  def track_action
+  def track_ahoy_visit
     ahoy.track 'Ran action', request.path_parameters
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,6 +16,8 @@ class ApplicationController < ActionController::Base
 
   layout 'layouts/main_site'
 
+  after_action :track_action
+
   protected
 
   # Strong params config for Devise
@@ -71,6 +73,11 @@ class ApplicationController < ActionController::Base
     return admin_path if resource.can? :view_admin_area
 
     user_profile_path( resource.username )
+  end
+
+  # Track all actions with Ahoy
+  def track_action
+    ahoy.track 'Ran action', request.path_parameters
   end
 
   def set_view_paths
--- a/config/initializers/ahoy.rb
+++ b/config/initializers/ahoy.rb
@@ -1,0 +1,5 @@
+class Ahoy::Store < Ahoy::DatabaseStore
+end
+
+# set to true for JavaScript tracking
+Ahoy.api = false
--- a/config/initializers/ahoy.rb
+++ b/config/initializers/ahoy.rb
@@ -1,5 +1,23 @@
+# Initialiser for Ahoy web stats; picks up various privacy-related settings
+# from ENV which are potentially useful for complying with GDPR.
+
+# Set AHOY_LINK_USERS to the string 'false' to disable automatic linking of
+# visits and users
 class Ahoy::Store < Ahoy::DatabaseStore
+  def authenticate( data )
+    # :nocov:
+    super unless ENV['AHOY_LINK_USERS'].presence == 'false'
+    # :nocov:
+  end
 end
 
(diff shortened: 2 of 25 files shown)


app/controllers/application_controller.rb
```diff
@@ -16,6 +16,8 @@ class ApplicationController < ActionController::Base
 
   layout 'layouts/main_site'
 
+  after_action :track_ahoy_visit
```
eliotsykes: I'm a little apprehensive about having tracking enabled as the default for (most) controllers. It feels like an insecure default, making it more likely a sensitive controller action introduced in the distant future has its parameters logged by mistake.

Consider disabling tracking by default and only enabling on a per-controller, per-action basis, e.g.

```rb
class SomeController
  enable_tracking only: [:some_action, :another_action]
end
```

Although this is more verbose, it is the more secure default. I'd avoid having an `except:` option so developers aren't tempted to safe-list entire controllers.

Also consider disabling tracking for all logged-in admin users, regardless of whether they are using the admin backend or the public-facing site.

You may also want to sanitize the parameters that are tracked by scrubbing any params that match the names configured in `config/initializers/filter_parameter_logging.rb`.

One more thing - are any senstive tokens logged currently? E.g. password reset tokens.

app/controllers/application_controller.rb
```diff
@@ -73,6 +75,11 @@ def after_sign_in_path_for( resource )
     user_profile_path( resource.username )
   end
 
+  # Track all actions with Ahoy
+  def track_ahoy_visit
+    ahoy.track 'Ran action', request.path_parameters
```
eliotsykes: Will this log all form submissions and their data? E.g. comment text, emails, etc. Consider if that's desirable, this could be recording a lot of undesirable data / sensitive PII / spam.

config/initializers/ahoy.rb
```diff
@@ -0,0 +1,23 @@
...
+  def authenticate( data )
+    # :nocov:
+    super unless ENV['AHOY_LINK_USERS'].presence == 'false'
+    # :nocov:
+  end
+end
+
+# Set AHOY_MASK_IPS to the string 'true' to mask IP addresses
+# (sets the last part of the address to zeros before storing it)
+Ahoy.mask_ips = true if ENV['AHOY_MASK_IPS'].presence == 'true'
```
eliotsykes: Can be reduced to:

```diff
-Ahoy.mask_ips = true if ENV['AHOY_MASK_IPS'].presence == 'true'
+Ahoy.mask_ips = ( ENV['AHOY_MASK_IPS'].presence == 'true' )
```

config/initializers/ahoy.rb
```diff
@@ -0,0 +1,23 @@
...
+
+# Set AHOY_COOKIES to the string 'false' to use anonymity sets instead
+# https://privacypatterns.org/patterns/Anonymity-set
+Ahoy.cookies = false if ENV['AHOY_COOKIES'].presence == 'false'
```
eliotsykes: Can be reduced as above.

app/controllers/application_controller.rb
eliotsykes: Consider limiting to only GET requests.

eliotsykes:
Is geocoding enabled by default? If so, consider disabling in some/all environments. It will either be creating geocoding jobs on the queue, or worse doing it synchronously on every tracked request:

```rb
Ahoy.geocode = false
```

app/controllers/application_controller.rb
denny: Iirc, I'm using a gem that you suggested to prevent the latter happening in production.

I wonder if I could record stats for all users, but have two views of that data - an 'end user activity only' for current_user.can?( :list, :stats ) and an 'all activity' view for current_user.can?( :list, :admin_stats ).

(The latter 'super stats admin' people should be able to view the end-user-only stats too, that's probably a useful filter to be able to choose even if you're not restricted to it.)

I was googling for options for views of this data last night and Blazer seems like the thing that gets paired with ahoy. Are there any others I should consider / any good articles I should read? I looked at chartkick but it seem similar-ish to Blazer and I couldn't find any docs for pairing it with Ahoy at first search.

eliotsykes: > Iirc, I'm using a gem that you suggested to prevent the latter happening in production.

I can't recall which gem this is. Something to discuss today perhaps.

> I was googling for options for views of this data last night and Blazer seems like the thing that gets paired with ahoy. Are there any others I should consider / any good articles I should read? I looked at chartkick but it seem similar-ish to Blazer and I couldn't find any docs for pairing it with Ahoy at first search.

Blazer's a good choice (one of its dependencies is chartkick).

PR #229, opened 2020-03-13
Working Akismet code and neater reCAPTCHA code

denny:


eliotsykes:
@denny Something I should have covered yesterday was Rails' built-in secrets handling.

In brief, Rails supports storing secrets in an encrypted file at `config/credentials.yml.enc`, and/or an unencrypted file at `config/secrets.yml` (both can be used together when needed).

`config/credentials.yml.enc` doesn't support ENV vars, so `config/secrets.yml` would be the choice here, as it allows you to reference ENV vars in the YAML like `<%= ENV["SOME_API_KEY"] %>`.

Secrets stored in `config/secrets.yml` are then accessible through `Rails.application.secrets`, e.g. `Rails.application.secrets.akismet_api_key`.

This can then be stubbed in a test similar to what we've been doing already:

```rb
allow(Rails.application.secrets).to receive(:akismet_api_key).and_return('test-key')
```

If you're interested in using `secrets.yml`, then create a file at `config/secrets.yml` with contents like below (adapted from https://github.com/rails/rails/blob/5-1-stable/railties/lib/rails/generators/rails/app/templates/config/secrets.yml)

- Replace values in file below `A-VERY-LONG-RANDOM-STRING` and `ANOTHER-VERY-LONG-RANDOM-STRING` with values generated by running `bin/rails secret`.

```yaml
# Be sure to restart your server when you modify this file.

# Your secret key is used for verifying the integrity of signed cookies.
# If you change this key, all old signed cookies will become invalid!

# Make sure the secret is at least 30 characters and all random,
# no regular words or you'll be exposed to dictionary attacks.
# You can use `rails secret` to generate a secure secret key.

# Make sure the secrets in this file are kept private
# if you're sharing your code publicly.

# Shared secrets are available across all environments.

shared:
  akismet_api_key: <%= ENV["AKISMET_API_KEY"] %>

# Environmental secrets are only available for that specific environment.

development:
  secret_key_base: A-VERY-LONG-RANDOM-STRING

test:
  secret_key_base: ANOTHER-VERY-LONG-RANDOM-STRING

# Do not keep production secrets in the unencrypted secrets file.
# Instead, either read values from the environment.
# Or, use `bin/rails secrets:setup` to configure encrypted secrets
# and move the `production:` environment over there.

production:
  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
```

Once you restart the app, you should be able to access `Rails.application.secrets.akismet_api_key`.

PR #230, opened 2020-03-14
Refactor tests to use seeds

denny:
This got a bit more involved than I had realised it would. On the bright side, I quite like the way the refactored Site Settings and Feature Flags stuff turned out.


spec/rails_helper.rb
```diff
@@ -36,9 +37,15 @@
...
 
-InsertSet.create! if InsertSet.first.blank?
-
 RSpec.configure do |config|
+  config.before( :suite ) do
+    # Wipe the test database, then load the required seed data
+    # (feature flags, capabilities and their categories, etc)
+    DatabaseCleaner.clean_with :truncation
+    ShinyCMS::Application.load_tasks
+    Rake::Task['db:seed'].invoke
```
eliotsykes: Loading the rake tasks may be adding a few seconds to the test run. If it is noticeable, consider requiring `seeds.rb` which may be enough to achieve the same:

```diff
-     ShinyCMS::Application.load_tasks
-     Rake::Task['db:seed'].invoke
+     require_relative '../db/seeds'
```

PR #232, opened 2020-03-15
Heroku-related tweaks and fixes

denny:
* Move dotenv into production area of Gemfile (rake tasks need it)
* Remove geocoding support from Ahoy stats for now - it's broken, something about licensing
* Disable forgery protection in production - not sure if this is controversial or not? It false positives a lot at work, fills our logs with crap. It seems to cope badly with proxies? It doesn't like my slightly gaffer-taped Heroku+Cloudflare SSL config, anyway.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,6 +1,9 @@
 Rails.application.configure do
   # Settings here will take precedence over those in config/application.rb
 
+  # The forgery protection seems to false positive a lot around proxies...
+  config.action_controller.forgery_protection_origin_check = false
+
   # Code is not reloaded between requests.
   config.cache_classes = true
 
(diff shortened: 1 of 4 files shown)


config/environments/production.rb
```diff
@@ -1,6 +1,9 @@
 Rails.application.configure do
   # Settings here will take precedence over those in config/application.rb
 
+  # The forgery protection seems to false positive a lot around proxies...
+  config.action_controller.forgery_protection_origin_check = false
```
eliotsykes: > Disable forgery protection in production - not sure if this is controversial or not? 

This line does *not* disable forgery protection which would be bad. It only disables the origin check within the forgery protection check, which is OK.

The forgery protection token is still checked on all form submissions, which is good enough.

For discussion and related code on this see the PR that introduced this config. There's a mention of proxies: https://github.com/rails/rails/pull/22263

PR #242, opened 2020-03-18
Spam comment moderation in admin area

denny:
Bit rough around the edges, but functional enough to be worth merging as a milestone.

Not (yet) including feeding spam confirmation or ham corrections back to Akismet.


app/controllers/admin/comments_controller.rb
```diff
@@ -0,0 +1,135 @@
...
+      flash[ :alert ] = t( '.spam_or_ham' )
+    end
+
+    redirect_to action: :index
+  end
+
+  def hide
+    authorise @comment
+    @comment.hide
+    redirect_to request.referer || comment_in_context_path( @comment )
```
eliotsykes: Consider using `redirect_back` instead - this handles the referer check. https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_back

app/helpers/main_site_helper.rb
```diff
@@ -36,6 +36,26 @@ def user_profile_link( user = current_user )
     link_to user_display_name( user ), user_profile_path( user.username )
   end
 
+  # Returns the path to a comment's parent resource, anchored to the comment
+  # Tries to fall back gracefully if comment is deleted/marked as spam
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/MethodLength
+  def comment_in_context_path( comment )
```
eliotsykes: The responsibility for knowing how to generate a comment's URL belongs to the comment itself more than anything else. For this reason, consider moving some/all this URL generation logic into a method on the `Comment` model. This would also help with reuse in the specs mentioned in the TODO comment below.

app/models/user.rb
```diff
@@ -68,8 +68,8 @@ def can?( capability_name, category_name = :general )
     cc = CapabilityCategory.find_by( name: category_name.to_s )
     return true if capabilities.exists? name: capability_name.to_s, category: cc
 
-    Rails.logger.debug  "Capability check failed: '#{username}' " \
-                        "cannot '#{capability_name}' '#{category_name}'"
+    Rails.logger.debug  'Capability check failed: ' \
```
eliotsykes: All models (and mailers and controllers) have a `#logger` method that returns `Rails.logger`. The `Rails.` can be dropped from line 71. https://guides.rubyonrails.org/debugging_rails_applications.html#sending-messages

spec/requests/admin/comments_spec.rb
```diff
@@ -0,0 +1,194 @@
+require 'rails_helper'
+
+RSpec.describe 'Comment moderation', type: :request do
+  before :each do
+    @admin = create :discussion_admin
```
eliotsykes: It doesn't look like `@admin` needs to be an instance variable. May be the case with some of these other `@instance_variables` defined in the `before` block.

PR #256, opened 2020-03-27
Newsletter models

denny:
Pulling this into master at this point not because it does anything useful yet, but because it doesn't break anything and I like keeping divergence to a minimum :)


app/models/concerns/email.rb
```diff
@@ -0,0 +1,19 @@
+# Common behaviour for models that store email addresses
+module Email
+  extend ActiveSupport::Concern
+
+  included do
+    validates :email, presence: true, uniqueness: true, case_sensitive: false
+    validates_with EmailAddress::ActiveRecordValidator
+    validates :canonical_email, presence: true
+    validates_with EmailAddress::ActiveRecordValidator
```
eliotsykes: Lines 7 and 9 are repeated, it looks like they should make use of the `field` or `fields` option to specify which attribute to validate: https://github.com/afair/email_address#rails-validator

app/models/do_not_contact.rb
```diff
@@ -0,0 +1,28 @@
+# Store details of people who have asked you not to contact them at all
+# Uses EmailAddress.redact to store/find addresses as a one-way hash
+class DoNotContact < ApplicationRecord
+  validates :email, presence: true, uniqueness: true
+
+  before_save :redact_email
+
+  def self.include?( email )
+    find_by( email: canonicalise_and_redact( email ) ) ? true : false
```
eliotsykes: This can be reduced using `exists?`:

```diff
-find_by( email: canonicalise_and_redact( email ) ) ? true : false
+exists?( email: canonicalise_and_redact( email ) )
```

db/migrate/20200310005657_create_subscriptions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSubscriptions < ActiveRecord::Migration[6.0]
+  def change
+    create_table :subscriptions do |t|
+      t.integer :list_id, null: false
+      t.integer :subscriber_id, null: false
+      t.string :subscriber_type, null: false, default: 'EmailRecipient'
+
```
eliotsykes: Consider introducing a unique composite index across these 3 columns to mitigate duplicate subscriptions.

app/models/subscription.rb
```diff
@@ -0,0 +1,10 @@
+# Model for mailing list subscriptions
+class Subscription < ApplicationRecord
+  validates :subscriber, presence: true
```
eliotsykes: Consider introducing a uniqueness validation to mitigate subscribers duplicating their subscription on the same list. See the `:scope` option for the uniqueness validator for more info.

PR #275, opened 2020-04-18
Add magic comment to freeze string literals

007lva:
I saw that [some files already have the magic comment](https://github.com/denny/ShinyCMS-ruby/search?q=%23+frozen_string_literal%3A+true&unscoped_q=%23+frozen_string_literal%3A+true) so I thought would be good to extend it because string literals are planned to be [frozen by default in Ruby 3.x](https://bugs.ruby-lang.org/issues/11473).


denny:
Thanks :)

PR #276, opened 2020-04-19
Email tracking (powered by Ahoy::Email)

denny:


app/controllers/admin/email_stats_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+class Admin::EmailStatsController < AdminController
+  before_action :set_ahoy_user
+
+  def index
+    page_num = params[ :page ] || 1
+    messages = Ahoy::Message
+    messages = messages.where( user: @ahoy_user ) if @ahoy_user
+    @messages = messages.order( 'sent_at desc' ).page( page_num )
+
+    authorise Ahoy::Message
```
eliotsykes, line 20: Although there are no db writes yet here, if there were, this delayed authorisation could allow unauthorised writes to the db.

Consider moving the `authorise Ahoy::Message` to the first line of this `index` action, to mitigate this and minimise the work done for an unauthorised user.

app/controllers/admin/web_stats_controller.rb
```diff
@@ -9,20 +9,23 @@
...
     visits = Ahoy::Visit
-    if params[:user_id].present?
-      @user = User.find( params[:user_id] )
-      visits = visits.where( user: @user )
-    end
+    visits = visits.where( user: @ahoy_user ) if @ahoy_user
     @visits = visits.order( 'started_at desc' ).page( page_num )
-    authorise @visits
+
+    authorise Ahoy::Visit
```
eliotsykes, line 20: As above, consider throwing out unauthorised users asap by moving `authorise Ahoy::Visit` to the first line of the action.

app/mailers/application_mailer.rb
```diff
@@ -4,6 +4,8 @@ class ApplicationMailer < ActionMailer::Base
 
   before_action :set_view_paths
 
+  track open: true, click: true
```
eliotsykes, line 7: For privacy friendliness, consider making these configurable settings and defaulting to `false`.

app/views/admin/menu/_menu.html.erb
```diff
@@ -26,7 +26,9 @@
   <%= render partial: 'admin/menu/menu_other' %>
   <% end %>
 
-  <% if current_user.can? :list, :web_stats %>
+  <% if current_user.can?( :view_web,   :stats ) ||
```
eliotsykes, line 29: To consolidate these `current_user.can?` checks into as few templates as possible, and so this responsibility isn't spread so much across templates, consider moving the checks from `_menu` to the top of the `_menu_stats` partial and doing an early `return` if they're unauthorised:

```erb
<% return unless current_user.can?(:view_web, :stats) || ... || ... %>
```

config/routes.rb
```diff
@@ -165,8 +168,12 @@
...
     mount LetterOpenerWeb::Engine, at: 'letter-opener' if Rails.env.development?
 
-    # The Ultimate Catch-All Route! Passes through to page handler,
-    # so that we can have top-level pages - /foo instead of /pages/foo
-    get '*path', to: 'pages#show'
+    # This catch-all route passes through to the Pages controller, allowing
+    # sites to have top-level pages (e.g. /foo instead of /pages/foo).
+    # The constraint here is to avoid also catching the open and click tracking
+    # routes which are appended by the Ahoy::Email engine.
+    get '*path', to: 'pages#show', constraints: lambda { |request|
```
eliotsykes, line 175: Likely a needless time sink, but there's *potentially* (no guarantee!) a workaround to this Ahoy catch all route issue involving these:

- Catch all route workaround with `railties_order` config: https://github.com/ankane/ahoy_email/issues/1
- And/or using the `Ahoy.api = false` setting to disable auto-mounting, so Ahoy::Email can be manually mounted in your routes.rb. https://github.com/ankane/ahoy_email/blob/b478b61cd4c8ee6d8f0bef3894cbfc5d32c5669c/config/routes.rb#L2

app/views/admin/menu/_menu.html.erb
denny: Ooh, I didn't realise you could return early from templates. That's handy.

PR #278, opened 2020-04-19
Email previews

denny:
Powered by RailsEmailPreview, this allows admins to look at the emails the site will send out under various conditions (user registration, comment notifications, etc).

More to do on this, but enough to merge at this point.


app/mailer_previews/discussion_mailer_preview.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  private
+
+  def fetch_comment
+    @comment_id ? Comment.find( @comment_id ) : mock_comment
+  end
+
+  def mock_comment
+    # TODO: Use factory?
```
eliotsykes, line 22: For now I'd try to avoid bringing in the extra overhead (although probably small) of loading the factorybot + faker gems into production just for this.

It also avoids the risk of calling a factory that unintentionally persists some data to the production db.

E.g. Calling `FactoryBot.build_stubbed(:comment)` or even just `FactoryBot.attributes_for(:comment)` will usually be harmless, but its possible the comments factory gets configured to save a User or something else to the db on these calls, so saving test data in the prod. db.

app/mailers/application_mailer.rb
```diff
@@ -1,17 +1,26 @@
-# Application Mailer
+# Application mailer base class (NB: user mailer does not inherit from this)
```
eliotsykes, line 1: > Application mailer base class (NB: user mailer does not inherit from this)

(Apologies if you're already aware of this and its just not desirable.) 

Devise's mailer's parent be configured to be `ApplicationMailer` using the `parent_mailer` config. See the related code here:

https://github.com/heartcombo/devise/blob/64863513506f4f2913510a050194d158c979eacc/app/mailers/devise/mailer.rb#L4

config/initializers/devise.rb
```diff
@@ -23,6 +23,7 @@
 
   # Configure the class responsible to send e-mails.
   # config.mailer = 'Devise::Mailer'
+  config.mailer = 'UserMailer'
 
   # Configure the parent class responsible to send e-mails.
   # config.parent_mailer = 'ActionMailer::Base'
```
eliotsykes, line 29: Here's the `parent_mailer` config.

app/mailers/application_mailer.rb
denny: Ah! No, I was cobbling bits together from a few different blog posts and none of them mentioned that. Although ...

config/initializers/devise.rb
denny: ... I had changed the line directly above it, without noticing it. Embarrassing. :laughing:

PR #279, opened 2020-04-19
Better tracking of email recipients with Ahoy

denny:


app/mailers/application_mailer.rb
```diff
@@ -17,6 +17,16 @@ def default_from_address
...
 
+  def ahoy_user( email_address, name = nil )
+    @user = User.find_by( email: email_address )
+    return if @user.present?
+
+    @user = EmailRecipient.find_by( email: email_address )
+    return if @user.present?
+
+    @user = EmailRecipient.create!( name: name, email: email_address )
+  end
```
eliotsykes, line 28: The diff below doesn't do much for me, but its a good opportunity to show pairing activerecord's `create_with` and `find_or_create_by(!)` methods: https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-find_or_create_by

```diff
-    @user = EmailRecipient.find_by( email: email_address )
-    return if @user.present?

-    @user = EmailRecipient.create!( name: name, email: email_address )
+    @user = EmailRecipient.create_with(name: name).find_or_create_by!(email: email_address)
```

app/mailers/discussion_mailer.rb
```diff
@@ -10,6 +10,8 @@ def parent_comment_notification( comment )
 
     subject = parent_comment_notification_subject( @reply )
 
+    ahoy_user( @parent.notification_email, @parent.author_name_any )
```
eliotsykes, line 13: Consider making it more explicit what work the `ahoy_user` method is doing, currently reading this here it appears a little mysterious. It might benefit from:

- possibly renaming `@user` to something generic that doesn't imply this is a `User` object. Perhaps `@recipient`, `@receiver`, or ?
- only doing the find-create work, and leave setting the `@user / @receiver` instance variable to the caller

```diff
-ahoy_user( @parent.notification_email, @parent.author_name_any )
+@receiver = find_or_create_receiver(@parent.notification_email, @parent.author_name_any)
```

```rb
def find_or_create_receiver(email_address, name = nil)
  User.find_by(email: email_address) || EmailRecipient.create_with(name: name).find_or_create_by!(email: email_address)
end
```

app/models/discussion.rb
```diff
@@ -9,10 +9,14 @@ class Discussion < ApplicationRecord
 
   # Instance methods
 
+  def notifiable?
+    resource&.user&.email&.present? ? true : false
```
eliotsykes, line 13: Remove the last safe-navigation operator to avoid the ternary expression:

```diff
-resource&.user&.email&.present? ? true : false 
+resource&.user&.email.present?
```

PR #284, opened 2020-04-20
Refactoring some parts to be more idiomatic code

007lva:
`Model.all.first` is the same than `Model.first` and the methods which return a boolean normally ends with a question mark.


denny:
Thanks :)

PR #293, opened 2020-04-23
Report spam/ham to Akismet, to train their model

denny:


spec/requests/discussions_spec.rb
```diff
@@ -256,6 +256,33 @@
       expect( response.body ).not_to have_css '.notices', text: I18n.t( 'discussions.add_comment.success' )
       expect( response.body ).not_to have_css 'h2', text: title
     end
+
+    it "doesn't save a new comment if Akismet classifies it as 'blatant' spam" do
+      skip 'Valid Akismet API KEY required' if ENV[ 'AKISMET_API_KEY' ].blank?
```
eliotsykes, line 261: Consider introducing VCR here to record the Akismet API call, and so the test can always be run and remove the `skip`.

To mitigate recording the Akismet API key in the VCR cassette file, configure VCR with its `filter_sensitive_data` method to replace the secret key with a placeholder:

```rb
VCR.configure do |config|
  config.filter_sensitive_data('<AKISMET_API_KEY>') { ENV['AKISMET_API_KEY'] }
end
```

PR #303, opened 2020-04-26
Neater auth checks in admin menu templates

denny:


docs/TODO.md
```diff
@@ -5,18 +5,20 @@
...
   * https://github.com/tinfoil/devise-two-factor
+* Configurable (per-site and per-user) menu order in admin area
 * In admin area, load the full dataset for user capabilities, feature flags,
-	and possibly site settings, and stick them in a hash somewhere - the menu
-	makes a crazy amount of hits on the db and I haven't even implemented half
-	of the 'feature hidden by flag/capability/etc' conditionals yet.
+  and possibly site settings, and stick them in some hashes, with some helper
+  methods to check them. The menu already makes a crazy amount of hits on the
+  db and there are still a load of 'feature hidden by flag/capability/etc'
+  conditionals to implement.
```
eliotsykes, line 16: > The menu already makes a crazy amount of hits on the db and there are still a load of 'feature hidden by flag/capability/etc' conditionals to implement.

For duplicated queries, this will be partly mitigated by Rails' built-in query cache, so this may not be such a big deal as we'd thought.

The query cache is scoped per-request. This means that a repeated `SELECT` query will only hit the DB once per request - with the result stored in the query cache. Subsequent repetitions of that query, on that request, will return the cached result.

If you're interested in seeing which queries are using the cache, tail the log and look for log lines containing SQL SELECT queries that are prefixed with `CACHE`. These SQL queries do not reach the DB, their results are being grabbed from the cache.

(Note there are some things that cause the query cache to be cleared during a request for good reason - such as an `UPDATE` statement.)

PR #305, opened 2020-04-26
Small improvements to Mailers

denny:


app/mailers/application_mailer.rb
```diff
@@ -1,32 +1,36 @@
...
-# Application mailer base class (NB: user mailer does not inherit from this)
+# Base class for Mailers
 class ApplicationMailer < ActionMailer::Base
   include FeatureFlagsHelper
 
   before_action :set_view_paths
   before_action :set_site_name
 
-  track open: true, click: true
+  track open: -> { track_opens? }, click: -> { track_clicks? }
```
eliotsykes, line 10: Consider disabling click-tracking for any emails that contain sensitive URLs such as password reset emails. 

The password reset URLs contain a secret token that should have minimal exposure and only be stored in as few places as possible, so try to avoid storing it in Ahoy's tables. 

Similarly if Ahoy stores the whole message body in its table(s) for password reset emails, try to avoid that.

PR #351, opened 2020-06-06
Pass read-only models to main site views

denny:
Minimal protection against accidentally shooting oneself in the foot in a main site template.


app/controllers/news_controller.rb
```diff
@@ -15,19 +15,19 @@ class NewsController < ApplicationController
 
   def index
     page_num = params[:page] || 1
-    @posts = NewsPost.recent_posts.page( page_num )
+    @posts = NewsPost.readonly.recent_posts.page( page_num )
```
eliotsykes, line 18: The `readonly` call could be applied throughout the controllers using pundit's policy scopes. See Pundit's README for usage on `policy_scope`. 

I'm not convinced it would be an improvement over what is already here, in fact it may make this clear code harder to follow, as well as introduce policies to the public-facing (non-admin) controllers which doesn't feel necessary yet.

Anyway something to consider. The code would become something like this if using pundit's policy_scope method:

```diff
-@posts = NewsPost.readonly.recent_posts.page( page_num )
+@posts = policy_scope(NewsPost).recent_posts.page( page_num )
```

PR #354, opened 2020-06-06
Add page listing recently-active discussions

denny:


app/controllers/discussions_controller.rb
```diff
@@ -19,8 +19,10 @@ class DiscussionsController < ApplicationController
   before_action :stash_comment,    except: %i[ index show add_comment ]
 
   def index
-    # TODO: list of recently-active discussions etc
-    redirect_to root_path
+    count = ( params[ :count ] || 10 ).to_i
+    days  = ( params[ :days  ] || 7  ).to_i
```
eliotsykes, line 23: Consider applying a sensible upper limit for `count` and `days` to mitigate users trying to query excessive amount of data in a request.

spec/factories/discussions.rb
```diff
@@ -3,5 +3,34 @@
...
   factory :discussion do
     hidden { false }
+
+    trait :with_0to4_comments do
+      after :create do |discussion|
+        create_list :comment, rand(5), discussion: discussion
+      end
+    end
+
+    trait :with_four_comments do
```
eliotsykes, line 13: Some of these factories look like they could be consolidated using factorybot's transient attributes. Consider introducing a transient attribute for `comments_count`.

PR #369, opened 2020-06-12
Refactor slug-related concerns and validations

denny:
Covering more of the codebase now, with a relatively consistent approach across different sorts of slug scope.


app/validators/unique_in_collection_validator.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# Validator for attributes that only need to be unique within a collection
+# e.g. page.slug within its section, blog_post.slug within the month of posting
+class UniqueInCollectionValidator < ActiveRecord::Validations::UniquenessValidator
+  def validate_each( record, attribute, value )
+    @record = record
```
eliotsykes, line 7: Using instance variables here may not be thread-safe. I *think* validator objects are reused across model objects. We can investigate further on a call to see. In the meantime, consider removing the `@record`, `@attribute`, and `@existing_error_count` instance variables and passing local variables `record`, `attribute`, and `value` to the methods.

app/validators/unique_in_collection_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+  def error_count
+    @record.errors[ @attribute ].size
+  end
+
+  def globally_unique?
+    @existing_error_count == error_count
+  end
+
+  def unique_in_collection?( value )
+    options[ :collection ].call( @record ).where( "#{@attribute} = ?", value ).blank?
```
eliotsykes, line 29: This looks like it'd work OK for saving a new model instance, but what about updates to an existing/persisted model instance when its slug does not change. Would it conflict on the slug?

```rb
p = NewsPost.first

# Does this raise validation error with slug?
p.update!(body: 'body updated')
```

In this similar validator, the id of the persisted model instance is used in the query, I think, to avoid this issue:

```rb
dupes = dupes.where("id != ?", record.id) if record.persisted?
```
Source: https://github.com/discourse/discourse/blob/70e50a2255ebb30c78891a7f846e833ab85b0a8b/lib/validators/unique_among_validator.rb#L16

PR #395, opened 2020-06-28
Group class-methods

007lva:
Just grouping some class methods(in classes with more than one)


denny:
Hi - thanks for your contribution, but I'm not going to merge this one, sorry. I prefer the self.foo naming approach myself (easier to see at a glance if the method I'm looking at is a class method that way) and it's also what [Rubocop's best practices guide](https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md#classes) suggests.

PR #400, opened 2020-07-02
Separate votes controller, anonymous votes

denny:


app/controllers/votes_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  def votable_models
+    Rails.application.eager_load! if Rails.env.development?
+    ApplicationRecord.descendants.select( &:votable? ).map( &:to_s )
+  end
+
+  def find_voter
+    @voter = current_user if user_signed_in?
+    return if @voter.present?
+
+    votable_ip = VotableIP.find_or_create_by!( ip_address: request.ip )
```
eliotsykes, line 57: To improve the performance of this query, consider adding an index to `votable_ips.ip_address`. Consider making it a unique index if there shouldn't be duplicates.

app/models/votable_ip.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+# Track IP address of anonymous votable voters ('likes' feature)
```
eliotsykes, line 3: This comment suggests an alternative (more intention revealing?) name for this model instead of `VotableIP` would be something like `Voter`, `AnonVoter`, `Guest`, `GuestUser`, etc.

denny: Yeah... my place of work made me strangely reluctant to tie the keyword 'Voter' into a 'likes' feature too tightly, it feels like too 'high value' a generic term for it to be used here :) I did have AnonVoter for a while, but same problem (albeit less so) really.

Also a slight niggle that two votes from an IP may not be from the same voter. I had VotingIP briefly but that's so close to using the gem name that I decided to do that in the end.

PR #433, opened 2020-08-02
Form handlers plugin

denny:
I think this has reached a mergeable milestone. :open_mouth: 

Still need to figure out a generic approach for each of the areas currently marked with FIXME comments near temporary hardcoded bodges (seed data, demo data, admin menus, admin breadcrumbs).

Also, need to decide at some point whether to aim towards extracting each plugin into its own repo, or keep them all inside the main one.


config/routes.rb
```diff
@@ -167,23 +171,34 @@
...
+    # LetterOpener provides a webmail UI for viewing sent emails (in dev only)
     mount LetterOpenerWeb::Engine, at: 'letter-opener' if Rails.env.development?
 
-    # This catch-all route passes through to the Pages controller, allowing
-    # sites to have top-level pages (e.g. /foo instead of /pages/foo).
+    ########################################
+    # ShinyCMS plugins
+
+    # ShinyForms: generic form handlers
+    mount ShinyForms::Engine, at: '/'
```
eliotsykes, line 196: Consider wrapping this with a conditional to check if the plugin is enabled, e.g.

```suggestion
    mount ShinyForms::Engine, at: '/' if defined?(ShinyForms)
```

Gemfile
```diff
@@ -3,8 +3,13 @@
 source 'https://rubygems.org' do
   gem 'rails', '~> 6.0.3'
 
+  # ShinyCMS plugins (comment out to disable)
+  gem 'shiny_forms', path: 'plugins/ShinyForms'
```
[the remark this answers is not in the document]
denny: Heh; I think there's a vestigial SHINY_FORMS_PLUGIN=1 kicking around in docs/env-sample from where I thought of this at some point and then got distracted :D Although a list is neater anyway. Cheers :)

plugins/ShinyForms/app/models/shiny_forms/form.rb
```diff
@@ -6,20 +6,27 @@ class Form < ApplicationRecord
...
-        Rails.logger.warn "Unknown form handler '#{handler}' (form ID: #{id})"
-        false
-      end
+    def false_after_logging_warning
+      Rails.logger.warn "Unknown form handler '#{handler}' (form ID: #{id})"
+      false
+    end
+
+    def handler_exists?
+      handlers.respond_to?( handler.to_sym )
```
eliotsykes, line 25: Could this allow an attacker to call one of the public methods `FormHandler` inherits from `Object` to be called?

If so, see the available public methods by running this in the rails console:

```rb
puts ShinyForm::FormHandler.new.public_methods
#puts Object.new.public_methods
```

Consider making this stricter by using `ShinyForm::FormHandler::FORM_HANDLERS ` as an allowlist.

denny: Good call, cheers. I didn't have the FORM_HANDLERS constant when I wrote this bit, I added that quite late in (and a bit reluctantly) because I couldn't see a neater way to generate the drop-down in the admin UI - unless I'd used the locale file as the data-source for there, which seems like a dodgy reversal of purpose. Now that I've got it though, definitely makes sense to use it here too.

denny:
@eliotsykes Have you got any thoughts for me on more generic ways to approach the issues I've flagged with FIXME comments?

Two in the admin templates (menu and breadcrumb links, which are increasingly striking me as so closely related as to probably deserve some conceptual merging or common helper methods or something):
https://github.com/denny/ShinyCMS-ruby/blob/18828b867f209368de7736e5b26cc1c8f06453f7/app/views/admin/includes/_breadcrumbs.html.erb#L29
https://github.com/denny/ShinyCMS-ruby/blob/18828b867f209368de7736e5b26cc1c8f06453f7/app/views/admin/menu/_menu.html.erb#L7

And two in seed/demo data:
https://github.com/denny/ShinyCMS-ruby/blob/18828b867f209368de7736e5b26cc1c8f06453f7/db/seeds.rb#L309
https://github.com/denny/ShinyCMS-ruby/blob/18828b867f209368de7736e5b26cc1c8f06453f7/lib/tasks/demo.rake#L18

... actually it's just occurred to me that I might be able to modify the approach used in the Gemfile for some/all of these. I'll give that a whirl now, but if you can see a neater approach do let me know :slightly_smiling_face:

eliotsykes:
For the admin menu, one option to consider: https://github.com/denny/ShinyCMS-ruby/commit/650cab945cc79123f3a22287ff607606f4eba6b1

PR #436, opened 2020-08-02
Add ShowHide concern, rename hidden* flags on various models

denny:
... still work to do here - this could be a polymorphic thing with more flexibility - but this is a small improvement for now.

PS: Wow this touched a lot of code, for a conceptually simple bit of refactoring!


app/models/concerns/shiny_show_hide.rb
```diff
@@ -0,0 +1,34 @@
...
+# Common behaviour for URL slugs
+module ShinyShowHide
+  extend ActiveSupport::Concern
+
+  included do
+    # Validations
+
+    # rubocop:disable Style/RedundantSelf
+    validates :show_on_site,  inclusion: { in: [ true, false ] }
+    validates :show_in_menus, inclusion: { in: [ true, false ] } if self.respond_to?( :show_in_menus )
```
eliotsykes, line 12: I'm not clear if this is doing what's intended. What's the purpose of the `if self.respond_to?(:show_in_menus)` check?

db/migrate/20200802201259_rename_hidden_flags.rb
```diff
@@ -0,0 +1,73 @@
+class RenameHiddenFlags < ActiveRecord::Migration[6.0]
+  def change
+    rename_column :blog_posts, :hidden, :show_on_site
+    change_column_default :blog_posts, :show_on_site, from: false, to: true
+    t = BlogPost.where( show_on_site: true  )
+    f = BlogPost.where( show_on_site: false )
+    t.update( show_on_site: false )
+    f.update( show_on_site: true  )
```
eliotsykes, line 8: Are you able to check if these migration lines are doing what's intended?

I've a feeling the following will happen:

```rb
# t and f are lazy relations, the queries will not run until the `update_all` calls.
t = BlogPost.where( show_on_site: true  )
f = BlogPost.where( show_on_site: false )

# NB. Assuming these updates should use `update_all` instead of `update`.

# 1. Runs `UPDATE blog_posts SET show_on_site = 'f' WHERE show_on_site = 't';
t.update_all( show_on_site: false )

# 2. At this point all records in blog_post now have show_on_site == false

# 3. Runs `UPDATE blog_posts SET show_on_site = 't' WHERE show_on_site = 'f';
f.update_all( show_on_site: true  )

# 4. Now all blog_posts have show_on_site == true
```

app/models/concerns/shiny_show_hide.rb
denny: Not everything that uses the ShowHide concern has a show_in_menus attribute.

db/migrate/20200802201259_rename_hidden_flags.rb
denny: Argh. Yes, you're correct I'm sure - I keep forgetting about the lazy thing :) How can I nail down the items collected in t and f before I run the updates?

app/models/concerns/shiny_show_hide.rb
denny: My intention is to refactor this more extensively at some point; it seems like an obvious candidate for a polymorphic trait that can store the show/hide status for any instance of any object in a number of named scenarios (so I get something like section.pages.show_on( :menus).all at the far end). That's a way down the to-do list while I'm trying to focus on forms though, so this is a bit of interim tidying that took much less time to think about :slightly_smiling_face:

eliotsykes: > Not everything that uses the ShowHide concern has a show_in_menus attribute.

Because of the scope of this line 12, this may not work as intended.

The scope here means `self` will be the class only (e.g. `Blog`), and not an instance of the class (e.g. `Blog.new`).

If line 12 isn't working as intended, this may work instead (needs testing):

```diff
-validates :show_in_menus, inclusion: { in: [ true, false ] } if self.respond_to?( :show_in_menus )
+validates :show_in_menus, inclusion: { in: [ true, false ] } if self.column_names.include?( 'show_in_menus' )
```

db/migrate/20200802201259_rename_hidden_flags.rb
eliotsykes: Try running a single update statement that flips the values using the `NOT` operator:

```rb
execute "UPDATE blog_posts SET show_on_site = NOT show_on_site"
```

PR #438, opened 2020-08-03
Add wrapper method for setting flash notice and redirecting

denny:
(Golfing meaningless code-quality metrics, tra la la)

### DIFF
--- a/app/helpers/admin_area_helper.rb
+++ b/app/helpers/admin_area_helper.rb
@@ -23,6 +23,11 @@ module AdminAreaHelper
         %w[ posts pages templates ].include?( controller_name ) )
   end
 
+  def redirect_with_notice( redirect_path, notice_message )
+    flash[ :notice ] = notice_message
+    redirect_to redirect_path
+  end
+
   def redirect_with_alert( redirect_path, alert_message )
     skip_authorization
     flash[ :alert ] = alert_message
--- a/app/controllers/admin/blog/posts_controller.rb
+++ b/app/controllers/admin/blog/posts_controller.rb
@@ -32,8 +32,7 @@ class Admin::Blog::PostsController < AdminController
     authorise @post
 
     if @post.save
-      flash[ :notice ] = t( '.success' )
-      redirect_to action: :edit, blog_id: @blog.id, id: @post.id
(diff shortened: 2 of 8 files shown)


app/helpers/admin_area_helper.rb
```diff
@@ -23,6 +23,11 @@ def html_editor_needed?
         %w[ posts pages templates ].include?( controller_name ) )
   end
 
+  def redirect_with_notice( redirect_path, notice_message )
+    flash[ :notice ] = notice_message
+    redirect_to redirect_path
+  end
```
eliotsykes, line 29: `redirect_to` can also accept `:notice`, `:alert`, and `:flash` arguments to modify the flash. See these examples in the docs:


> It is also possible to assign a flash message as part of the redirection. There are two special accessors for the commonly used flash names alert and notice as well as a general purpose flash bucket.


```rb
redirect_to post_url(@post), alert: "Watch it, mister!"
redirect_to post_url(@post), status: :found, notice: "Pay attention to the road"
redirect_to post_url(@post), status: 301, flash: { updated_post_id: @post.id }
redirect_to({ action: 'atom' }, alert: "Something serious happened")
```

https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to

denny: Yeah, that's what I tried first; it seems to add the message as a GET param, meaning my URLs looked scruffy and all my tests broke, so I rolled my own version that didn't do that. Was I misusing the real one?

eliotsykes: This the `redirect_to` method signature:

```rb
redirect_to(options = {}, response_options = {})
```

The 1st hash arg, `options`, would be the same options you'd pass to `url_for` to generate a URL.

The 2nd hash arg, `response_options`, would have the `:status`, `:flash`, `:notice`, `:alert` key-value pairs.

When it didn't work correctly, the `options` and `response_options` may not have been separate hash arguments.

Try to ensure they're separate hash args, something like the below, and see how that goes:

```diff
-flash[ :notice ] = t( '.success' )
-redirect_to action: :edit, blog_id: @blog.id, id: @post.id
+redirect_to({ action: :edit, blog_id: @blog.id, id: @post.id }, notice: t('.success'))
```

PR #454, opened 2020-08-06
Generic handling of plugin admin menus and breadcrumbs, seed data, and demo data

denny:


denny:
@eliotsykes Okay, on a scale of 1 to 'you are a bad person', how does this look? :stuck_out_tongue_winking_eye:

eliotsykes:
> how does this look? 

I'm just catching up with the last few PRs, so far so good 👍

PR #458, opened 2020-08-07
Convert news section feature to a plugin

denny:
Hopefully this is a fairly clean implementation of a plugin, to use as a jump-off point for:

(a) Thinking about writing a plugin generator (see notes in dev docs folder)

(b) Looking at how much work would be involved in shipping the plugins separately (as gems or whatever)

This plugin is slightly less tied into the CMS than the Forms plugin, but definitely not fully independent - there's still some inheriting from main app base classes, as well as relying on the rest of the CMS rather than a 'dummy' app for its tests.

(Also a bunch of other features rely on the News feature for their tests; comments/discussions particularly.)


denny:
Closing in favour of separate PRs:
* #459 (remove news feature)
* #460 (refactor plugin stuff)
* #461 (add news plugin)

PR #467, opened 2020-08-09
ShinyNews plugin: adds a news section to ShinyCMS

denny:
Hopefully this is a fairly clean implementation of a plugin, to use as a jump-off point for:

(a) Writing a plugin generator

(b) Converting the rest of the existing features to plugins where appropriate

(c) Looking at how much work would be involved in shipping the plugins separately


app/models/user.rb
```diff
@@ -28,6 +28,7 @@ class User < ApplicationRecord
   # Admin content: throw an error if it hasn't been removed or reassigned
   has_many :blogs,      dependent: :restrict_with_error
   has_many :blog_posts, dependent: :restrict_with_error
+  has_many :shiny_news_posts, dependent: :restrict_with_error, class_name: 'ShinyNews::Post'
```
eliotsykes, line 31: I suspect there is a way to define this news_posts association from the plugin, e.g. in its `engine.rb`, you could call something like:

```rb
User.has_many :shiny_news_posts, dependent: :restrict_with_error, class_name: 'ShinyNews::Post'
```

This would reduce coupling, though I'm not sure how desirable that is though, because when a dev looks at `app/models/user.rb` source, they won't see the association defined like you do now on line 31, so you're losing some discoverability.

denny: Oh, neat. I think it's worth the obfuscation for the decoupling, although I think I've found a slightly more discoverable place to stash that line of code anyway... the news post model: https://github.com/denny/ShinyCMS-ruby/commit/b37c8c3ddf1646c74599b763004406a655b9af07

eliotsykes: Looks great 👍

PR #485, opened 2020-08-13
Profile pages plugin

denny:
This set of features is quite heavily entangled with Devise user account stuff, but this is a first pass at separating some of it out a bit.


app/helpers/admin_area_helper.rb
```diff
@@ -25,8 +25,15 @@ def html_editor_needed?
...
-    ::Plugin.loaded.map( &:underscore )
+  def plugins_for_admin_menu
+    plugins = []
+    ::Plugin.loaded.each do |camelcase|
+      underscored = camelcase.underscore
+      if File.exist? Rails.root.join "plugins/#{camelcase}/app/views/#{underscored}/admin/menu/_section.html.erb"
+        plugins << underscored
+      end
+    end
+    plugins
```
eliotsykes, line 36: Perhaps `::Plugin.loaded` should return an array of `Plugin` or `ShinyPlugin` objects so the objects can have useful behaviour attached, to allow something like:

```rb
def plugins_for_admin_menu
  plugins = []
  ::Plugin.loaded.each do |plugin|
    plugins << plugin if plugin.in_admin_menu?
  end
  plugins.map(&:underscored_name)
end
```

PR #494, opened 2020-08-15
Plugin <==> main app hooks for tags, menus, and admin toolbar

denny:
Also, ShinyNews pundit policy for news posts no longer inherits from main app.

Plus odds and ends of tidying around the edges.

### DIFF
--- a/app/models/plugin.rb
+++ b/app/models/plugin.rb
@@ -38,4 +38,14 @@ class Plugin
   def self.capability_category_to_plugin_name( capability_category )
     capability_category.to_s.sub( %r{_[a-z]+$}, '' ).classify.pluralize
   end
+
+  def self.base_records
+    return @base_records if @base_records
+
+    @base_records = []
+    loaded.each do |plugin|
+      base_records << plugin.constantize::ApplicationRecord
+    end
+    @base_records
+  end
 end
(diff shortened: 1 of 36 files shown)


plugins/ShinyNews/app/models/shiny_news/post.rb
```diff
@@ -34,13 +33,17 @@ def path( anchor: nil )
 
     # Specify policy class for Pundit
     def policy_class
-      Admin::PostPolicy
+      ShinyNews::PostPolicy
```
eliotsykes, line 36: Consider delegating to class method:

```diff
-ShinyNews::PostPolicy
+self.class.policy_class
```

app/models/plugin.rb
```diff
@@ -38,4 +38,14 @@ def self.loaded
   def self.capability_category_to_plugin_name( capability_category )
     capability_category.to_s.sub( %r{_[a-z]+$}, '' ).classify.pluralize
   end
+
+  def self.base_records
+    return @base_records if @base_records
+
+    @base_records = []
+    loaded.each do |plugin|
+      base_records << plugin.constantize::ApplicationRecord
```
eliotsykes, line 47: Is `base_records` call on line 47 correct? Looks like it'll call the `.base_records` method recursively.

PR #497, opened 2020-08-17
Pass Plugin objects around more instead of just their names

denny:
And add some more useful methods to the objects.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -106,8 +106,8 @@ class ApplicationController < ActionController::Base
     prepend_view_path 'app/views/shinycms'
 
     # Add the default templates directory for any loaded plugins above that
-    Plugin.loaded.each do |plugin_name|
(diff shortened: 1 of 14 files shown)


app/controllers/application_controller.rb
```diff
@@ -106,8 +106,8 @@ def set_view_paths
     prepend_view_path 'app/views/shinycms'
 
     # Add the default templates directory for any loaded plugins above that
-    Plugin.loaded.each do |plugin_name|
-      prepend_view_path "plugins/#{plugin_name}/app/views/#{plugin_name.underscore}"
+    Plugin.loaded.each do |plugin|
+      prepend_view_path "plugins/#{plugin.name}/app/views/#{plugin.name.underscore}"
```
eliotsykes, line 110: Consider extracting the views path `"plugins/#{plugin.name}/app/views/#{plugin.name.underscore}"` in to a method of `Plugin`

PR #501, opened 2020-08-17
Change ShinyPost.recent_posts method to be a scope instead

denny:


app/models/concerns/shiny_post.rb
```diff
@@ -35,6 +35,7 @@ module ShinyPost
 
     scope :not_future_dated, -> { where( 'posted_at <= ?', Time.zone.now.iso8601 ) }
     scope :published,        -> { visible.merge( not_future_dated ) }
+    scope :recent,           -> { order( posted_at: :desc ) }
```
eliotsykes, line 38: It looks like the `recent` scope can return unpublished posts. If so, perhaps the `published` scope needs to be used somewhere.

PR #611, opened 2020-10-03
Fix user mailers

denny:
Add _instructions to a bunch of method names.


denny:
These send now, but they have blank bodies, which is ... suboptimal.

PR #633, opened 2020-10-07
Upgrade Rails stack from 6.0.3.3 to 6.0.3.4 due to CVE-2020-8264

denny:


denny:
Not sure why CodeCov has skipped the Algolia search code here; I haven't hit their API rate limit for the month, and it runs locally with 100% coverage.

denny:
Re-running CI on the branch to see if the issue is transient.

denny:
Yep, that fixed it. Go figure.

PR #635, opened 2020-10-08
Remove namespacing within plugin controllers, models, and helpers

denny:
Checking to see if the fully-qualified class names in some of these files are the reason Codebeat has been downgrading me lately...


denny:
Didn't seem to make much difference to code quality metrics, so closing this without merging.

PR #666, opened 2020-10-21
Fix kaminari pagination in plugin engines

### LINKED ISSUE #626: Paging is broken on some admin pages
I think it's any page with more than one part path after /admin - e.g. /admin/news will page okay, but /admin/newsletters/editions won't.

It looks like the pager block (pulled in from a partial) is trying to render with the URL /admin/editions, dropping the middle part of the path.

Possibly an easy fix, but I couldn't see it at first glance.

(Paging is powered by kaminari, not will_paginate)

denny:
Kaminari has a very long-standing bug that causes it to explode messily when used with Rails engines, which for some reason they seem averse to fixing:
https://github.com/kaminari/kaminari/pull/322
https://github.com/kaminari/kaminari/pull/636

Fortunately for me (and my CMS made almost entirely out of Rails engines), @cbeer released kaminari_route_prefix as a workaround:
https://github.com/cbeer/kaminari_route_prefix

This PR fixes #626


denny:
@eliotsykes Thought you might be interested in this one; turns out that Kaminari is fundamentally broken when used with Rails engines :open_mouth:

Fortunately there is a workaround.

PR #668, opened 2020-10-22
Routes and route concerns for pagination and admin search features

denny:


denny:
Hakiri fail was their code exploding, rather than my code having a security issue.

PR #676, opened 2020-10-26
Use ActiveStorage to handle image elements in templated content

denny:
TODO: Update demo data


denny:
This code breaks in multiple places if ActiveStorage is set to use local disk instead of S3. :boom:

PR #717, opened 2020-11-13
Rubocop GitHub code-scanning workflow

denny:
Opening this PR so I can close it with a note to remind myself that this failed to fail when I provoked it, so something's not quite right with it currently.

Also it was SLOW AF - it actually took longer to just run rubocop here on GitHub than my entire test suite takes on CircleCI or Travis CI.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -1,0 +1,31 @@
+name: "Rubocop"
+
+on: push
+
+jobs:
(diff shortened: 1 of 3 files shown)


denny:
See top note; I branched from this and made a breaking change (updated rubocop to a version that didn't like the current code), and it broke at CircleCI but still passed here, so ... that seems suboptimal.

PR #764, opened 2020-12-05
Refactor theme files

denny:
Move theme view templates, images, and stylesheets, into a new top-level `themes` directory.


denny:
Codefactor issues are all style complaints about CSS that isn't mine.

PR #802, opened 2020-12-25
Refactor: Use Enumerable#map

007lva:
Use [Enumerable#map](https://ruby-doc.org/core-2.7.2/Enumerable.html#method-i-map) instead of Enumerable#each + mutable Array

### DIFF
--- a/app/models/concerns/shiny_html_template.rb
+++ b/app/models/concerns/shiny_html_template.rb
@@ -37,13 +37,8 @@ module ShinyHTMLTemplate
     def self.available_templates
       return [] unless template_dir
 
-      template_names = []
-
       filenames = Dir.glob '*.html.erb', base: template_dir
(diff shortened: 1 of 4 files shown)


denny (changes requested):
Hey, this looks really nice, thanks very much!

I think this will fail rubocop as it stands, because I have that set to prefer .collect to .map - can you just check that it passes rubocop with the included .rubocop.yml (as well as the test suite obviously) and if not, make any changes that requires? Then I can get this merged in 🙂

denny:
I wanted to get this in before the 21.1 release tomorrow, so I pulled it across into another branch and fixed up the rubocop stuff, and I'm going to merge it from there - hope that's okay with you, I think it should still show you as the author of the commits. Thanks again for the contribution, this is much nicer than my code! 🙂

PR #890, opened 2021-01-28
Allow users to add and remove links on their profile page

denny:


denny:
Some quite possibly hideous JavaScript going on here, but it seems to work for the most part.

PR #929, opened 2021-02-16
Move (almost) all remaining main app functionality into a plugin

denny:
This is a first pass at moving all of the remaining main_app functionality into a new `ShinyCMS` core plugin, with an eye to eventually making it possible to mount whichever bits of ShinyCMS you want to use, inside your own existing Rails application.

The main/container app has been renamed to `ShinyHostApp`. Theoretically its only job now is to mount the ShinyCMS core and feature plugins.

In practice, it also still has to mount some Rails engines from gems that were not immediately co-operative about being mounted inside the new ShinyCMS core plugin, and currently it still contains all the config files.

Things that aren't done yet:
* Move most of the config files
* Move the rest of the Rails engines provided by gems
* Do a name-based refactoring to get rid of the 'Shiny' prefix on the many classes and modules that now have a 'ShinyCMS::' namespace anyway (e.g. `ShinyCMS::ShinyPlugin` can become `ShinyCMS::Plugin`)
* Update documentation!
* ... probably some other stuff I can't think of right now :)

So, it's not exactly finished yet, but it passes the tests (including some new ones) and I figured **merge early, merge often** :D


denny:
Best Typo-CI warning ever! 🤣 

> "thankyouverymuch" is a typo. Did you mean "¯\_(ツ)_/¯"?**

denny:
@eliotsykes GitHub has cruelly suggested you as a reviewer for this PR 😀 

(Don't review it, it's huge! But in case you are interested, here's the first draft of the core plugin, as discussed in today's emails.) 😅

PR #959, opened 2021-02-27
Rename Shiny* model concerns

denny:
I think this is the last big batch of renaming to get things that went from being `Shiny{Thing}` to being `ShinyCMS::Shiny{Thing}` after the core plugin refactor, to now be just `ShinyCMS::{Thing}`. Because much as I love a good 'Shiny', the duplication in every class name was a bit OTT. 🙂


denny:
shiny

PR #977, opened 2021-03-03
Import routes from 'route partials'

denny:
This PR adds a utility method which can be used to import individual routes or groups of routes into a routes file, from a separate file or files.

It also includes some 'pre-rolled' route partials (mostly shipped within the core ShinyCMS plugin, plus two in ShinyPages) which can be imported into the host app's config/routes.rb with a few simple lines - to easily load the various parts of ShinyCMS into any application. This is demonstrated in the ShinyHostApp routes file.

There are also some route partials which modularise the routes file in the core plugin, and one for admin routing concerns which is used by almost all of the feature plugins.


denny:
@eliotsykes If you have any time free, I'd be interested to hear what you think of this one at some point...

I originally thought route partials would be nice while I was doing the core plugin refactor - a way to give people easy integration hooks for bolting ShinyCMS into an existing Rails app - but I couldn't work out how to make it happen. Then I ran across a clue a few days ago while I was looking for something completely different 🙂 

So I'm very pleased with the general concept and how it works in practice, but I'm less sure about the implementation details. Specifically, the eval-ish thing in the monkey patch - they're generally frowned upon for being lorry-sized security holes, aiui 🤔

denny:
Selected highlights, for your clicky convenience...

One of the simpler plugin routes.rb files, which now imports a partial to get the admin routing concerns: https://github.com/denny/ShinyCMS-ruby/pull/977/files#diff-c88826fb906803d7b45f059671f6f01813d7a9ecd0380d3cf9cdfad0ea49796c

The main app routes.rb, which imports several partials to hook everything together and hopefully demonstrate the integration thing: https://github.com/denny/ShinyCMS-ruby/pull/977/files#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5

A couple of partials...
`admin_route_concerns` (imported by most plugins): https://github.com/denny/ShinyCMS-ruby/pull/977/files#diff-87099734873b391e866daca399e5c03af3f0aa6fbf2187742cf5e8d59a15603a
`top_level_pages` (imported at end of main_app routes.rb): https://github.com/denny/ShinyCMS-ruby/pull/977/files#diff-e19cb2ae146e780693c00c10701c4bf4bd7c6063316f1ae5e044651968cb516c

Here Be Monkeys; the `import_routes` method: https://github.com/denny/ShinyCMS-ruby/pull/977/files#diff-756bb7310a55b91907666d1c70792e7bc9128f4a0c8e8239889f0b6f96d8651c

eliotsykes:
It looks good to me. I don't think using instance_eval in this context is a
security concern, I can't see a way for a web-based attack to leverage it.

The only suggestion I have is if you wanted to improve the line numbers and
source code in errors while debugging. You can include the optional args to
instance_eval for filename and line numbers.

There's a brief blog post on this here:
https://olabini.se/blog/2008/01/ruby-antipattern-using-eval-without-positioning-information/

I think for import_routes, you could do something like:

```ruby
path = Rails.application.root.join(
"plugins/#{plugin}/config/routes/#{file}.rb" )
contents = File.read(path)
instance_eval(contents, path, lineno = 0)
```

PR #988, opened 2021-03-06
Refactor supporting structures for 'tools' - Blazer, REP, etc.

denny:


denny:
Hakiri is a false alarm, again. It's been literally YEARS since you were supposed to put `protect_from_forgery` in Rails controllers, ffs.

PR #1001, opened 2021-03-09
Use rails-lts fork of ActiveRecord::SessionStore with fix for CVE-2019-16782

denny:


denny:
Hakiri still picking up the CVE on the gem I've swapped out; safe to ignore, for now.

PR #1013, opened 2021-03-14
Tests for corner-cases, and various bikeshedding

denny:
This is the PR equivalent of a shed at the bottom of the garden. A few good bits in there, and then a fair bit of whatever seemed like a good idea at the time, apparently.

There are quite a few tests though, so that's nice.


denny:
Note To Self / passers-by; that thing I did in the plugin engines.rb files? Don't do that. 🤦🏼

PR #1014, opened 2021-03-14
FactoryBot was loaded conditionally for very good reason, i.e. it's not present in production.

denny:
🤦🏼 🤦🏼‍♀️ 🤦🏼‍♂️


denny:
At some point I may have to figure out why this manifested as only the tags admin concern test failing.

denny:
Also, props to BugSnag, who managed to send me an email telling me what the actual problem was as soon as it hit my staging server. CircleCI seem to be eating my test output lately, no doubt as a result of another piece of idiocy on my part.

PR #1047, opened 2021-04-05
Checking and enforcing boundaries, with Packwerk

denny:
This PR adds [Packwerk](https://github.com/shopify/packwerk#readme) and then follows that by moving various code into public interface areas, or building abstractions to present a public interface to private code, resulting in a project which now has only a dozen or so .

Although it wasn't done for this reason, it's been interesting to watch the project quality score (as calculated by Ruby Critic) climb through this bit of work. Modular code is clean code is good code; who knew? 😉


denny:
Marking this as Do Not Merge for now, as there's something weird about the coverage report:
```
  Files          350       318    -32     
  Lines         5092      4546   -546     
``

If anything this PR has added files (and, presumably, lines). I certainly don't think it shrank the project by 10%!

denny:
Merging this branch into clone of main seems to rediscover all the missing files; presumably just something caching related going on here (see 'two hard problems').

PR #1069, opened 2021-04-10
Refactor discussions controller

denny:
Split main site comment actions off into new comments controller (with more REST-ish action/method names), and move views etc to match.


denny:
Can't blame Typo CI for not liking that variable name. Wasn't keen myself. 😆

PR #1075, opened 2021-04-21
Better interception of Blazer's inheritance from host app ApplicationController

denny:
Thank you @eliotsykes for suggesting the approach taken here! I am much happier with this gank separated out from the host app's standard ApplicationController.


denny:
Typo-CI hit is kind of a fair cop here, but I'm leaving it in anyway. I am a bad person. 😉

PR #1088, opened 2021-04-29
Accessibility improvements for Halcyonic theme

denny:
* Alt text inputs for index template images
* Accessibly hidden labels for login, search, and newsletter subscribe forms


denny:
More to do here! Update template data, update demo data, check rest of templates and forms for similar issues, ...

PR #1130, opened 2021-05-23
Refactor: share common behaviour with modules rather than inheritance

denny:
This PR moves the common behaviour for controllers and components out of Base* classes and into modules that are then included where neccessary. This feels like a better way to reach across the plugin boundaries for this shared stuff.


denny:
I'm not happy with the naming scheme in this; is there an equivalent to the BaseController (etc) pattern for mixins?

Currently I am seriously tempted to use 'Stuff'... `ShinyCMS::ViewComponentStuff`, `ShinyCMS::AdminControllerStuff`, etc 😉

PR #1148, opened 2021-06-13
Remove Webpacker

denny:
The recent collection of security vulnerabilities caused by Webpacker depending on ancient versions of things, tied together with the constant bloody nightmare that trying to configure it has been throughout this project, culminated in my ripping it out tonight.

Weirdly this seemed to go very smoothly; removed it from package manager config files and reinstalled packages, then removed every file that had the word 'webpacker' in it (except one of my docs, which I updated), then jiggled a couple of jQuery things slightly, and ... that's it, I think.

So long, good riddance.


denny:
@eliotsykes This should probably count as a defeat, but it sure feels like a victory 😉

PR #1958, opened 2025-04-02
New plugin: ShinyShop.,]

denny:
Initial block of work on the payment handling part of the new ShinyShop plugin.

This first attempt uses Stripe to handle the payment stuff, and is probably more tightly-coupled to that than is ideal... contributions to support other payment processors would be _very_ welcome, jump into the comments or send me an email if you have thoughts.

I think GoCardless looks interesting - possibly cheaper for some end-use cases, and a different payment flow (underpinned by direct debits rather than credit cards), so should force me to think about where the dividing line is between generic payment processing code and specific support for a particular payment processor.


denny:
Finally got this back in shape and passing the code quality checks. It's still a bit threadbare, but probably safe to merge soon-ish (for next release?), and then keep working on making it more polished after that.

denny:
No idea why that just failed the coverage test after that merge, it has full coverage. I think there's a race condition in @GitHub's @CodeCov integration.

denny:
"WARN[0001] Conflict when uploading: A test report for commit 81feeefccb87d6457f343176f75795976117d782 already exists, skipping upload"

